Issue #6, opened 2018-11-19
Support HTTPS/WSS


kmlucy:
I have Dozzle reverse proxied with SSL. The main page loads, but the websocket connections do not:
```
Mixed Content: The page at 'https://URL/dozzle/container/ID' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://URL/dozzle/api/logs?id=ID'. This request has been blocked; this endpoint must be available over WSS.
```

I have websocket proxying setup in NGINX, but it seems like Dozzle is still looking for a ws:// connection. NGINX doesn't show any logs, so it is being blocked in the browser and not even making it to my proxy.

If I temporarily allow unsafe connections, Dozzle exits with this error:
```
2018/11/19 21:48:00 websocket: request origin not allowed by Upgrader.CheckOrigin
```

amir20:
I suspect I need to change `ws://` to `wss://`. I think I'll try to do this locally with Caddy server. See if that works. 

If that works, it might be another flag. This is an interesting setup you have. But I think it is common and something that should be supported. 

I may send you a test image instead because it will take a while to make this work locally.

amir20:
@kmlucy  I think I got it to work. I haven't used Nginx for a while so you'll have to convert whatever I did to Caddy. 

If I change `ws` to `wss` and use this proxy config, then it works. 


```

https://localhost:1234 {
    tls self_signed

    proxy / localhost:8080 {
        transparent
        websocket
    }
}
```

I can create a new flag called `-wss` (maybe) to enable wss instead. But I want to make sure the proxy idea works for you. According to the [documentation](https://caddyserver.com/docs/proxy), you have allow upgrade of websockets when doing proxy. 

Something like 

```
header_upstream Connection {>Connection}
header_upstream Upgrade {>Upgrade}
```

I am not sure what the equivalent of that is for Nginx. 

If that works for you, then I can get the flag working.

kmlucy:
@amir20 There is a similar configuration for NGINX and it is what I've been using. Were you able to get it to work with Caddy without changing Dozzle?

I'd like to try to --wss flag (or whatever the flag ends up being). I could try it out and see if it fixes the problem I'm having.

amir20:
@kmlucy Try with https://github.com/amir20/dozzle/releases/tag/v1.2.1. You need to use `--ssl` flag. You can do `docker run amir20/dozzle --help` if you want to see help. 

[...]

PR #8, opened 2018-11-26
Uses EventSource instead of websockets

amir20:


amir20:
@kmlucy I did some research and it seems like `Sever Events` is superior compared to `WebSockets`. I started playing with it and it does seem simpler. I also tested it behind a proxy and SSL is no longer issue. So I am going to merge this and up the minor version since it is a new feature. I am removing the `--ssl` flag in favor of just not needing a flag at all since `EventSource` is just over HTTP/s.

amir20:
Related to #6

amir20:
Also forgot to mention that auto reconnect with eventsource seems to be built out of the box for most browsers. 👍

kmlucy:
I just tested the latest version, and it seems to work fine. I removed the --ssl flag and removed the portions of my proxy location block relevant to websockets. 

With the new version, the logs do not auto-update reliably, however. They seem to only update every few minutes. Is that a bug, or will continuous updating not work with server events?

amir20:
> With the new version, the logs do not auto-update reliably, however. 

No that must be a bug. I was able to test it and it does work for me. I did a HTTP load test with `siege` and saw the logs fly by. 

Is there a way you can help me debug? I haven't checked the refresh when on a proxy. I can check that next. 

Let me know if you find anything.

PR #16, opened 2019-03-14
assets/pages: preserve whitespace of log messages

iwittkau:


iwittkau:
```
go get: error loading module requirements
The command "go get -u github.com/gobuffalo/packr/packr" failed and exited with 1 during .
```

amir20:
Hmm not sure why the broken build. I run it locally.

assets/pages/Container.vue
```diff
@@ -91,4 +91,8 @@ export default {
 .is-fullheight {
   min-height: 100vh;
 }
+
+.item-message {
```
amir20: There is already a class `.text`. Can't you use that? Since Vuejs has the scoped CSS, you don't need to worry about class collisions.

iwittkau: Changed it, thanks for explaining. I'm new to vue.

amir20:
Looks good. I'll merge but won't release until tomorrow when I get home.

PR #17, opened 2019-03-15
Revert "assets/pages: preserve whitespace of log messages"

amir20:
Reverts amir20/dozzle#16

@iwittkau I am going to have to revert these changes because it looks like it break page wrap. Which is something that should work. I should have asked for a screenshot before merging.


amir20:
Screen Shot 2019-03-14 at 7 20 48 PM

Merging because it looks to broken.

PR #40, opened 2019-06-14
 Add deploy with docker swarm 

ZEROF:


amir20:
Thanks. I’ll merge.. but I think you would want this on every node and not just manager.

PR #84, opened 2019-07-22
Container Restrictions & Run Docs

adamzero1:
- Allow the viewable containers to be filtered using regex on container name
- Steps to run locally


docker/client.go
```diff
@@ -40,6 +44,35 @@ func NewClient() Client {
...
 
+func (d *dockerClient) GetContainerById(containerId string) (Container, error) {
+    containerListOptions := types.ContainerListOptions{
+        Filters: filters.NewArgs(
+             filters.Arg("id", containerId),
+         ),
+    }
+    list, err := d.cli.ContainerList(context.Background(), containerListOptions)
+    if err != nil {
+        return Container{}, err
```
amir20: In a case of error, it would probably more sense to just return `nil`. Returning an empty struct makes more sense for arrays.

docker/client.go
```diff
@@ -40,6 +44,35 @@ func NewClient() Client {
...
+    }
+
+    if len(list) == 0 {
+        return Container{}, errors.New(fmt.Sprintf("Unable to find container with id: %s", containerId))
```
amir20: Same here

docker/client.go
```diff
@@ -40,6 +44,35 @@ func NewClient() Client {
 	return &dockerClient{cli}
 }
 
+func (d *dockerClient) GetContainerById(containerId string) (Container, error) {
+    containerListOptions := types.ContainerListOptions{
+        Filters: filters.NewArgs(
+             filters.Arg("id", containerId),
+         ),
+    }
+    list, err := d.cli.ContainerList(context.Background(), containerListOptions)
```
amir20: Would it make more sense to call `ListContainers()` instead and filter by id? That way the struct `Conatiner{}` doesn't have to be created again.

docker/client.go
```diff
@@ -40,6 +44,35 @@ func NewClient() Client {
 	return &dockerClient{cli}
 }
 
+func (d *dockerClient) GetContainerById(containerId string) (Container, error) {
```
amir20: Do you have to add a quick test for this? It should be pretty easy.

amir20:
Overall, looks great. Just a couple quick comments. I should be able to merge today if you can fix 'em.

docker/client.go
amir20: I was thinking about this on my way to work. Since you are already using filters. Would it make more sense to just send the name pattern to docker instead and filter within `ListContainers`?

adamzero1: When I did it this morning, thought it looked messier to have the additional logic in the `streamLogs` function, but you're prolly right.

adamzero1:
Last commit should hopefully resolve the issues, if you could have a look and let me know that would be grand :)

main.go
```diff
@@ -137,7 +143,19 @@ func (h *handler) listContainers(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	err = json.NewEncoder(w).Encode(containers)
+
+    filteredContainers := []docker.Container{}
+	for _, container := range containers {
```
amir20: I am confused. I thought here you would want to pass `nameFilter` to `ListContainers`? Sorry if I wasn't clear last time. I was thinking when you have `ListContainer(filter)` then you could filter all containers by name from docker and you don't need to a regex anymore? Is that too complicated? In other words, I am suggesting remove `containerRestrictions` in favor of letting docker to the filtering. This would be more clear for users of Dozzle who are familiar with Docker. The functionality of filter would mirror `docker ps`

amir20: I mean `containerRestrictions` as a regex. You still need a filterName or something to pass to docker

adamzero1:
Yeah I'm with you, hadn't thought of that tbh, as all the docs don't mention the filters support regex.
Having just tested it looks like it does.
Should now be using the `--filter` syntax to match docker
third time lucky! :P

amir20:
I think looks good. I brought your code over locally to test it. I'll merge soon.

amir20:
I wasn't sure why it wasn't running tests for you. Which is weird. I am new to Github actions. I created https://github.com/amir20/dozzle/pull/85 with some tests and updated to be called `filterName` instead. Thanks for your contribution!

amir20:
Closing in favor of https://github.com/amir20/dozzle/pull/85

PR #85, opened 2019-07-22
Adds tests and uses `filterName` instead. 

amir20:
@adamzero1 I moved some tests around. I also fixed some of the code so that the logic for finding the right container is push back to the `docker.Client`


README.md
```diff
@@ -66,7 +72,20 @@ Dozzle follows the [12-factor](https://12factor.net/) model. Configurations can
...
+| `--filterName` | `DOZZLE_FILTERNAME` | `""` |
 
 ## License
 
 [MIT](LICENSE)
+
+### Building
+To Build and test locally:
+
+1. Install NodeJs.
```
amir20: Also cleaned this up.

docker/client.go
```diff
@@ -27,21 +30,47 @@ type dockerProxy interface {
...
 type Client interface {
 	ListContainers() ([]Container, error)
+	FindContainer(string) (Container, error)
 	ContainerLogs(context.Context, string, int) (<-chan string, <-chan error)
 	Events(context.Context) (<-chan events.Message, <-chan error)
 }
 
 // NewClient creates a new instance of Client
-func NewClient() Client {
+func NewClient(filter string) Client {
```
amir20: Added this so that it always filters regardless of callers.

main.go
```diff
@@ -39,6 +40,7 @@ func init() {
 	pflag.String("base", "/", "base address of the application to mount")
 	pflag.String("level", "info", "logging level")
 	pflag.Int("tailSize", 300, "Tail size to use for initial container logs")
+	pflag.String("filterName", "", "Filters containers by name")
```
amir20: Used `filterName` instead to be similar to docker manual

amir20:
All tests pass. Going to merge.

PR #109, opened 2019-08-23
added --showAll flag that will show all containers (default: false)

m1ome:
Sometimes is quite handy to look upon `down` container logs, this PR make it possible.
It creates a `--showAll` flag, which by default set to `false`.

It will not break anything.


amir20:
This looks good. Thank you for your contribution @m1ome! 🎉

PR #419, opened 2020-04-30
Add sample Nginx config to README.md

whalehub:
This pull request adds a minimal sample Nginx config to the README.md file with the options that are required for getting Dozzle to work over HTTPS.


whalehub:
@amir20 This should do the trick.

amir20:
Thanks!

PR #527, opened 2020-06-22
added support for toggling between new light theme and dark theme.

catalinmunteanu:
Hello!

Please consider reviewing my PR which introduces an option to switch to a light-theme.

Implemented new settings toggle button which allows us to switch between the original Dozzle theme and a light-theme.
Screenshot 2020-06-22 at 08 33 11

Thank you so much,
Catalin MUNTEANU

### DIFF
--- a/assets/styles.scss
+++ b/assets/styles.scss
@@ -14,15 +14,19 @@ html {
   overflow-y: unset;
   scroll-snap-type: y proximity;
   background-color: #111;
+
+  &.has-light-theme {
+    background-color: #F8F8F8;
+  }
 }
 
 body {
   font-family: "Roboto", sans-serif;
   color: #ddd;
-}
 
-h1.title {
-  font-family: "Gafata", sans-serif;
+  html.has-light-theme & {
+    color: #333;
+  }
 }
 
 html.has-custom-scrollbars {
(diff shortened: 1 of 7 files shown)


amir20:
Thanks for this. Let me pull it locally and see how it looks.

amir20:
@catalinmunteanu can you take a look at the settings page? The lines are still white. 

Screen Shot 2020-06-22 at 11 37 29 AM

catalinmunteanu:
Yes. I will return with some minor tweaks. Thank you!

amir20:
It would be great if Dozzle automatically read user's settings. Something I have used is https://github.com/jloh/bulma-prefers-dark before which works pretty well. Just a thought...

catalinmunteanu:
@amir20, I optimized the styles a bit for the light theme and now all visible html elements are covered by the use-light-theme option. Thank you!
Screenshot 2020-06-22 at 22 39 48

amir20:
Still one more. 
Screen Shot 2020-06-22 at 2 51 33 PM

assets/styles.scss
```diff
@@ -71,3 +71,79 @@ html.has-custom-scrollbars {
...
+
+  body {
+    color: #333;
+  }
+
+  h1.title {
+    text-shadow: 1px 0px #ffdd57;
+  }
+
+  .has-underline {
```
amir20, line 86: Are you familiar with Vue? Can you move this to Vue component page instead? Something like 

```
.has-underline {
  border-bottom: 1px solid #fff;
  padding: 1em 0px;
  margin-bottom: 1em;

  html.has-light-theme & {
    border-bottom: 1px solid #DCDCDC;
  }
}

```

assets/styles.scss
```diff
@@ -71,3 +71,79 @@ html.has-custom-scrollbars {
...
+    border-bottom: 1px solid #DCDCDC;
+  }
+
+  code {
```
amir20, line 90: Same with this

assets/styles.scss
```diff
@@ -71,3 +71,79 @@ html.has-custom-scrollbars {
...
+    background-color: #999;
+    color: #FFF;
+  }
+
+  .splitpanes--vertical > .splitpanes__splitter {
```
amir20, line 95: Move this to `App.vue`

amir20:
Overall, it looks ok. There is still one page broken. But the styles for light are all in `styles.css`. This is a problem because as other styles on components change, they are disconnected from the light version. 

Can you move them to Vue components instead please?

assets/styles.scss
```diff
@@ -71,3 +71,79 @@ html.has-custom-scrollbars {
...
+html.has-light-theme {
+  background-color: #F8F8F8;
+
+  body {
+    color: #333;
+  }
+
+  h1.title {
+    text-shadow: 1px 0px #ffdd57;
+  }
```
amir20, line 84: These probably make sense here since they are also defined in this file. You can make it more `scss` like and use `&`.

catalinmunteanu:
Light Theme styles are now defined in each component 😃
Thank you

amir20:
Thanks. This looks good.

Issue #23, opened 2019-05-23
custom ports


vpotap:
for custom port usage with command 
$ docker run --volume=/var/run/docker.sock:/var/run/docker.sock -p 8888:1224 amir20/dozzle:latest --addr localhost:1224

you still get access to localhost:8080 due to hardcoded port in Dockerfile
ENTRYPOINT ["/dozzle"]
EXPOSE 8080

amir20:
That's a good point. Send me a PR and we can remove it from Dockerfile. I still haven't figured out what `EXPOSE` is good for...

alex-m-humai:
automatic expose for example via something like
traefik (https://traefik.io/) does not work any more without a defined port.
it is possible to specify manually but I think this should be in the dockerfile again.

amir20:
I also use traefik. I think if you are advanced enough to use traefik then setting up ports should be expected. Also, I remember reading in the traefik's manual that defining ports are favored as containers can expose multiple ports.

disklosr:
TLDR; `EXPOSE` should be added back, and contrary to what OP says, it doesn't cause any problem when using a custom port.

---

@amir20 did you try to reproduce @vpotap 's issue? I tried and couldn't, which wasn't surprising to me since EXPOSE` is only for documenting the default ports a container is expected to listen on, and while it does expose the port inside the container, it doesn't publish anything to the host.

So if you run:
```
$ docker run --volume=/var/run/docker.sock:/var/run/docker.sock -p 8888:1224 amir20/dozzle:latest --addr localhost:1224
```
Docker will only open 8888 on the host. If you try to access ```localhost:8080``` it shouldn't work. If it worked for op I suspect they did publish port 8080 manually (using -P switch) but didn't pay attention to it.

And so, there's no need to remove `EXPOSE` instruction from Dockerfile because it's just documenting the default port of the app inside the container so you know it by just reading the Dockerfile. But most importantly, this isn't incompatible with `--addr` switch.

Uses cases like Traefik where you do not need to publish any port should just work by default because they don't care about ports and will just use the default one, hence the need for EXPOSE instruction. The burden of adding extra config should be on people who need a custom port, and in this case, having an `EXPOSE` instruction won't do any harm and won't interfere with anything.

amir20:
[...]

PR #564, opened 2020-07-07
Add expose instruction to Dockerfile

disklosr:
Adding expose instruction to Dockerfile as its needed for service discovery by reverse proxy (like Traefik)


amir20:
Thanks for contribution @disklosr. `EXPOSE` was removed in https://github.com/amir20/dozzle/issues/23. But it makes sense to add it back. One problem is that with `--addr` you can overwrite the port. So the `EXPOSE` port wouldn't be accurate anymore. Any thoughts here? Should I deprecate `--addr`? I am not sure if it is even used by anybody.

amir20:
I won't be merging this right now. I will create an issue to deprecate `--addr`.

PR #832, opened 2020-11-19
Added choice of clock format to settings

dima-bzz:


amir20:
Thanks! I'll review it soon. But it does look like you removed the default behavior on using browser's locale. I'll have to keep that feature around and make it a "force 24 hour" style. 

Is there a reason I would want .babelrc PR vs this one?

dima-bzz:
In the babel branch, older browsers do not support formatToParts. So I put it in the settings. You don't need to do this in the master branch.

amir20:
Closed via 5f0cdee76ccdea6805cf4624019739e07591931d

PR #915, opened 2021-01-02
Tries to user readers instead of channels

amir20:


amir20:
This greatly increases memory footprint. From something like 12mb to 1.2mb.

PR #1584, opened 2021-10-29
Add clear logs functionality

### LINKED ISSUE #1547: Possibility to clear/hide old logs per container / globally
When working on my services a lot of times I want to see what logs are produced
by an action which I'm going to perform next (e.g., when I run some end to end test).

In order to quickly see what did get logged I often clear my console before I execute
such action. In dozzle it's not possible to do so and I have to remember/figure out
which log messages are 'old' and which are new - produced by the action.

I would love if there would be some way to clear/hide these old logs (e.g. by hitting some
button) - globally and/or per service.

PS: thanks for dozzle! It's really nice.

Coteh:
This is my first pass at adding clear logs functionality to dozzle. Fixes #1547.

image
(edit: updated screenshot to reflect recent changes from upstream)

I created a new toolbar that houses clear logs and download buttons. I spaced them out so that it's less likely user will accidentally click one button when they meant to click the other.

I also moved the search bar below the new toolbar so that actions - specifically, the Download action - can still be accessed.
Screenshot 2021-10-29 at 15-54-22 si-db - Dozzle

The scroll progress component's pointer events have also been turned off as well (via CSS `pointer-events: none`) so that the Download button can still be clicked when it appears.

I have picked Ctrl-L as the hotkey to clear logs, but if you prefer a different keybind let me know.

Bin icon is from [IcoMoon Free Pack](https://icomoon.io/#preview-free).

Issues/Caveats:
- Cannot scroll up to see prior logs when logs are cleared. This may be ok though, since the user intended to clear logs.
- Logs will come back if you click on another container then go back to the container you cleared logs on.
- Search component covers the scroll progress in this version. If I kept the search component where it was previously, it would cover the close button and part of the download button. Perhaps this is acceptable? Or do you prefer having the search bar where it was before? I have screenshotted both versions below:
image
image

- Ctrl-L to clear hotkey will not activate when search box is active. Is this acceptable, or would you like me to look into how to have the hotkey work even when the search box is active?

If you have any pointers on how I could resolve these issues or if there's any other comments, I would appreciate it.

(below is text I have written in my commit, some of it has already been mentioned but I have left this here for completeness)
Other Changes:
- Add actions toolbar for actions pertaining to container logs
- Move Download button to actions toolbar, remove text and margin to match clear logs button
- Move search box below actions toolbar
- Turn off pointer events for scroll progress component so that Download button can be clicked
- Adjust spacing of container bar to accomodate the new actions toolbar


amir20:
This is a very good start. Thank you for doing it. I am still reviewing it. But here are some high level feedback. I think the UI could be prettier. 

> I created a new toolbar that houses clear logs and download buttons. I spaced them out so that it's less likely user will accidentally click one button when they meant to click the other.

Did you see the prototype I put in the original bug? Here is the [screenshot](https://user-images.githubusercontent.com/260667/138946215-991f15f7-683e-4810-9e55-9ee989c9012f.jpg). I don't think two double headers look pretty. It takes away from vertical space and makes the title really fat. On mobile it would probably look worse. 

I recommend inventing a new bar that is floating in the middle top like the screenshot I had. I like how you brought the download button there. So that could be the two buttons for now. 

 > I have picked Ctrl-L as the hotkey to clear logs, but if you prefer a different keybind let me know.

This might be a problem because cmd/ctrl+l on most browsers is keyboard shortcut to address bar. It would be bad experience if you steal a default behavior. How about `shift+cmd+L`?

amir20:
This is a very good start. Thank you for doing it. I am still reviewing it. But here are some high level feedback. I think the UI could be prettier.

I created a new toolbar that houses clear logs and download buttons. I spaced them out so that it's less likely user will accidentally click one button when they meant to click the other.

Did you see the prototype I put in the original bug? Here is the [screenshot](https://user-images.githubusercontent.com/260667/138946215-991f15f7-683e-4810-9e55-9ee989c9012f.jpg). I don't think two double headers look pretty. It takes away from vertical space and makes the title really fat. On mobile it would probably look worse.

I recommend inventing a new bar that is floating in the middle top like the screenshot I had. I like how you brought the download button there. So that could be the two buttons for now.

I have picked Ctrl-L as the hotkey to clear logs, but if you prefer a different keybind let me know.

This might be a problem because cmd/ctrl+l on most browsers is keyboard shortcut to address bar. It would be bad experience if you steal a default behavior. How about`shift+cmd+L`?

> Cannot scroll up to see prior logs when logs are cleared. This may be ok though, since the user intended to clear logs.

That's totally fine. It was user's intent. 

> Logs will come back if you click on another container then go back to the container you cleared logs on.

I think this is fine too. A workaround would be to store the timestamp when the log was cleared and then send that to the API to only fetch logs _after_ the timestamp. Too much work IMO. 

> Search component covers the scroll progress in this version. If I kept the search component where it was previously, it would cover the close button and part of the download button. Perhaps this is acceptable

I think is ugly. But if you do the hover toolbar like I suggested above, this should go away.

assets/components/LogEventSource.vue
```diff
@@ -30,6 +31,12 @@ export default {
     this.flushBuffer = debounce(this.flushNow, 250, { maxWait: 1000 });
     this.loadLogs();
   },
+  mounted() {
+    hotkeys("command+l, ctrl+l", (event, handler) => {
```
amir20, line 35: What would happen if there are multiple panes open?

assets/components/LogContainer.vue
```diff
@@ -1,14 +1,26 @@
...
         </div>
-        <div class="column is-narrow">
+        <div class="column is-clipped is-paddingless"></div>
+        <div class="column is-narrow is-paddingless" v-if="closable">
+          <button class="delete is-medium" @click="$emit('close')"></button>
+        </div>
+      </div>
+      <div class="mr-0 is-vcentered is-hidden-mobile">
+        <div class="is-flex">
+          <a @click="onClearClicked" class="button is-small is-outlined" id="clear">
```
amir20, line 18: Since there is only an icon, a tooltip with `title=` could be good.

assets/components/LogContainer.vue
```diff
@@ -73,6 +81,11 @@ export default {
       return config.base;
     },
   },
+  methods: {
```
amir20, line 84: I recommend putting the keyboard shortcut here to be closer to the button.

assets/components/LogEventSource.vue
amir20: Like mentioned, maybe adding `shift` would better so the default behavior is not cancelled.

assets/components/ScrollProgress.vue
```diff
@@ -90,6 +90,7 @@ export default {
 .scroll-progress {
   display: inline-block;
   position: relative;
+  pointer-events: none;
```
amir20, line 93: Good call

assets/components/LogContainer.vue
```diff
@@ -1,14 +1,26 @@
...
+        <div class="column is-clipped is-paddingless">
           <container-stat :stat="container.stat" :state="container.state"></container-stat>
         </div>
-        <div class="column is-narrow">
+        <div class="column is-clipped is-paddingless"></div>
+        <div class="column is-narrow is-paddingless" v-if="closable">
+          <button class="delete is-medium" @click="$emit('close')"></button>
+        </div>
+      </div>
+      <div class="mr-0 is-vcentered is-hidden-mobile">
```
amir20, line 16: Nice that you remembered to hide on mobile

amir20 (changes requested):

assets/components/LogContainer.vue
```diff
@@ -1,14 +1,26 @@
...
+        <div class="is-flex">
+          <a @click="onClearClicked" class="button is-small is-outlined" id="clear">
+            <span class="icon">
+              <icon name="bin"></icon>
+            </span>
+          </a>
+          <div class="is-flex-grow-1"></div>
```
amir20, line 23: So like I mentioned in my comment, having another toolbar makes the UI very clunky. I recommend looking in to centering a toolbar and using `position: fixed` to keep it floating.

Coteh:
> Did you see the prototype I put in the original bug? Here is the [screenshot](https://user-images.githubusercontent.com/260667/138946215-991f15f7-683e-4810-9e55-9ee989c9012f.jpg). I don't think two double headers look pretty. It takes away from vertical space and makes the title really fat. On mobile it would probably look worse.
> 
> I recommend inventing a new bar that is floating in the middle top like the screenshot I had. I like how you brought the download button there. So that could be the two buttons for now.

My apologies for not following the screenshot. I'll see if I can get that to work, but one issue I personally have with it is that I feel it kind of gets in the way of the logs on the top. I think it should slide in from the top when you hover your mouse anywhere on the top half of the log view. What do you think?

> On mobile it would probably look worse.

I don't think the toolbar is setup to appear on mobile atm. Do you think it should appear on mobile as well? If so, then the hover suggestion I mentioned above may not work for mobile specifically. Maybe the user could tap the top half of the log view for the toolbar to appear? Or it could just appear there at all times, like in original screenshot.

> This might be a problem because cmd/ctrl+l on most browsers is keyboard shortcut to address bar. It would be bad experience if you steal a default behavior. How about`shift+cmd+L`?

That works! 👍 

> I think is ugly. But if you do the hover toolbar like I suggested above, this should go away.

Sounds good, I can get to work on changing the toolbar. Again, sorry for not following that screenshot, and thanks for the valuable feedback! 😄

Coteh:
Another question, do you think the floating menu should be coloured white like in your screenshot, or are you open to a floating menu that is coloured black like the rest of the interface:

image

I think the way it was depicted in your screenshot makes it look more like a notification at first glance rather than a toolbar. But what do you think?

Also, I don't use the light theme personally, but I could make the toolbar white for the light theme though.

assets/components/LogEventSource.vue
Coteh: > What would happen if there are multiple panes open?

Good question. I just had a look and it appears the hotkey clears logs on multiple panes at once, while the button just clears logs for the pane it's on. I think we will need to keep track of the active pane when hotkey is pressed. What do you think?

amir20: That's good actually. With keyboard shortcut I had expected for only one to work, but I guess that makes sense. I think what it does now it makes sense. Keyboard to clear all, mouse click to do the clicked one.

amir20:
> My apologies for not following the screenshot. I'll see if I can get that to work, but one issue I personally have with it is that I feel it kind of gets in the way of the logs on the top. I think it should slide in from the top when you hover your mouse anywhere on the top half of the log view. What do you think?

I think whatever makes sense. I had actually thought of doing it on the side but thought it would be weird since there is already the scroll button.  If you do on the side then maybe the buttons should be stacked on top of each other. 

I actually really like the idea of only showing it when you move the mouse. But UX gurus might say that's confusing since users now need to do some action for a button to show up. Maybe don't do it now and first just to that absolute minimum and we can launch it. Collect feedback and then update it. 

> I don't think the toolbar is setup to appear on mobile atm. Do you think it should appear on mobile as well? If so, then the hover suggestion I mentioned above may not work for mobile specifically. Maybe the user could tap the top half of the log view for the toolbar to appear? Or it could just appear there at all times, like in original screenshot.

Let's just hide on mobile as you had. That makes sense to keep mobile as simple. No one is going to be downloading or clear logs on mobile. 

> Sounds good, I can get to work on changing the toolbar. Again, sorry for not following that screenshot, and thanks for the valuable feedback! 😄

Hey no worries. You are not getting paid for this :P so whatever time you can put it, it's great. If you become a regular contributor, I will list your name in the readme. 


> Another question, do you think the floating menu should be coloured white like in your screenshot, or are you open to a floating menu that is coloured black like the rest of the interface:

That was just a test. I think the color should match the background. Which should be `--body-background-color`. You can probably make it slightly opaque.  One thing which is weird in your screenshot is the border. Maybe match the icons closer to the icons on the left for consistency?


Screen Shot 2021-10-30 at 2 18 36 PM


> Doesn't look too bad with the black actually (IMO), but perhaps it could be coloured grey like the buttons on the left menu.

I think minus the weird white background on the button, it looks good. My comment above should work for both though. Matching the rounded icons on the left could be best option. 


Thanks again for doing all of this! At some point, when you get really close, we can merge and deploy `master` somewhere to play with it. 

One final idea, if this "component" becomes too complicated with logic then abstract out to a new component so it's not all mixed with the rest.

assets/components/LogContainer.vue
Coteh: Done. How does this look?

image

assets/components/LogContainer.vue
Coteh: Done.

assets/components/LogContainer.vue
Coteh: Used `position: absolute` to center it, which takes into account the side menu. Done.

Screenshot 2021-11-04 at 20-42-35 si-db - Dozzle

Coteh:
Here's the new toolbar, will fix merge conflict and you can review afterwards.

Screenshot 2021-11-04 at 20-47-54 si-db - Dozzle
Screenshot 2021-11-04 at 20-47-42 si-db - Dozzle

Tooltip:
image

Coteh:
Another change I have made is making the top bar paddingless. Here's before and after screenshots with two panes. What do you think? I think I prefer the top bar being more compact, which allows the toolbar to be moved up a bit further to the top.
Screenshot 2021-11-04 at 21-18-06 si-sessions - Dozzle
Screenshot 2021-11-04 at 21-18-30 si-sessions - Dozzle

amir20:
Looks really good. I wonder if the hover should be yellow. But to be honest, this is polished enough to merge. I may consider a slight white border to make the toolbar pop.

amir20:
I can't tell why, but the tests are never ran when you've sent me PRs. I wonder if I did something wrong with the settings to not work.

amir20:
`master` should be in docker hub now. I played with it a little. Fixed one minor bug. I'll keep trying it out until tomorrow and push an official release. Thanks for doing this!

PR #1604, opened 2021-11-19
(WIP) Move actions toolbar to side menu

### LINKED ISSUE #1597: New download and clear buttons hide log
Hey there! 

I really like the new clear button, but I think the positioning is not ideal - because it hovers above the log content it sometimes gets in the way.

For example, if there is important information in the first two lines it is pretty hard to read. Also selecting text behind these buttons is a little clumsy. This is a problem especially for devices with smaller screens, like tablets:

image

Coteh:
Fixes #1597.

image

I no longer have time to work on this issue until next week. Two things I could not get around to in time:
- When clicking on dropdown item, **both** dropdown items get an "is-active" class that cannot be removed. [The examples](https://oruga.io/components/Dropdown.html#examples) don't have this problem.
- I would like to vertically align the icons and the text. Again, I won't have time until next week to polish this off.

Your call whether to wait until next week when I have time, or if you want to fix them now.

Btw, thanks for setting things up with that comment. :)

### DIFF
--- a/assets/components.d.ts
+++ b/assets/components.d.ts
@@ -15,6 +15,7 @@ declare module 'vue' {
     LogEventSource: typeof import('./components/LogEventSource.vue')['default']
     LogViewer: typeof import('./components/LogViewer.vue')['default']
     LogViewerWithSource: typeof import('./components/LogViewerWithSource.vue')['default']
+    MdiDotsVertical: typeof import('~icons/mdi/dots-vertical')['default']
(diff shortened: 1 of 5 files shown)


amir20:
Hey thanks for doing this. I play around with it tomorrow and see if I can fix it. I also noticed the weird `is-active` bug. I think it might be because there is no `v-model` so Vue doesn't know what is selected. I have migrated to Oruga which is new. So I am sure there are bugs.

amir20:
I merged this locally. Thanks for your help.

PR #1639, opened 2022-01-16
Add jump to context (Work in progress)

### LINKED ISSUE #1586: Need help: Jump to context from search
One of the complains I have myself when searching is that search only shows lines with keyword. It would be helpful to see before and after lines. 

One thought is too have a little icon next to each line when searching to jump to context and highlight the line temporary to a different color. 

Implementation is a little tricky because each line needs to be brought in to view and selected. Possibly a new component would be suited.

Coteh:
Here's what I have so far for the new jump to context feature. Will close #1586 when it's done and merged.

Screenshot 2022-01-16 at 02-08-55 eager_pascal - Dozzle
Screenshot 2022-01-16 at 02-09-24 eager_pascal - Dozzle

To Do
- [ ] ~Find a way to notify ScrollableView component when user has clicked on jump to line button, so that the view won't stick to the bottom and prevent the view from scrolling to the line. (I added some timing hacks for now)~ edit: I have later found that there actually doesn't seem to be a difference whether the view sticks to the bottom or not.
- [x] Improve UI for jump to context button. Maybe position it a bit more, and/or add an outline around the line to give indicator of which line will be jumped to.
- [ ] ~Fix text coloring when there are colored lines~ edit: omitting for now, from initial investigation this will require editing all the default colors provided by [ansi-to-html](https://github.com/rburns/ansi-to-html) library. (using the `colors` constructor option) I think once this feature gets merged and into the next release, we can get a better idea of how these colors should be modified when highlighted.
- [x] Add light mode colors for selected line
- [x] Fix jump to context not working on split pane views
- [x] Fix horizontal scrollbar if log line exceeds width of page
- [ ] ~Fix regression where exiting search in split pane mode sometimes causes the pane on the right to offset out of view, requires scrolling up slightly to fix~ edit: found this bug outside of my branch so I'll be making a new issue for it soon


amir20:
I like what you got so far. I was thinking of a slightly different UX. What if you put the menu dot only on hover on left?

Screen Shot 2022-01-18 at 3 34 57 PM
 
Right now the mock looks very crowded. 

I'll wait for some more updates and review again.

Coteh:
Found some time tonight to make a few updates. Here's how it looks now.

Screen Shot 2022-01-24 at 12 20 18 AM

Still want to take a bit of time to clean things up and put the finishing touches in. I'll take the PR off of draft when it's fully ready. Feel free to give some more feedback regardless. :)

amir20:
This is perfect. It's exactly what I had in mind except the hover. Is that a tooltip? You can use the same menu as the top right?

Coteh:
Oh I see, you'd prefer that clicking on the three dots will open a context menu like the one already on the top right corner of the page, and there would be only one option at the moment which says "Jump to context" which when clicked will jump to the line like I already have. I can do that, makes sense to have another menu especially since we are re-using the three dots icon.

amir20:
Yes, that's what I was thinking. I was also thinking other options in the future like "Clear above here" but that could be done later.

amir20:
I see some changes were pushed. Should I review again?

Coteh:
Yeah feel free to take another look if you like. I just got it working on split panes. However, I still need to address some regressions. (they're in the checklist up top)

image

It looks like this now. You can hover your mouse over the three dots and it'll bring up the menu like this. (overlaps the three dots until you move your mouse away)

amir20:
@Coteh do you think you can finish all outstanding actions by March 15th? I am going to have some free time then so I can try to merge it.

amir20:
I also pulled latest and played with it a little. The spacing when search is off. But the functionality works great.

Coteh:
> @Coteh do you think you can finish all outstanding actions by March 15th? I am going to have some free time then so I can try to merge it.

I have other commitments next weekend but I'll try to pick these off one by one throughout the week.

> The spacing when search is off.

I had the chance to play around a bit with this today and I was able to get the spacing down to this size.

Screenshot 2022-03-06 at 16-44-18 objective_vaughan - Dozzle

I think this is probably the best we can go with the icon still being legible. The spacing also looks pretty close if not the same to how the spacing usually is when not searching. Let me know your thoughts.

Coteh:
Screenshot 2022-03-06 at 14-19-45 si-app - Dozzle

Also, some of the colors emitted by ansi-to-html don't look so great when highlighted by the context. I have found that this can possibly be fixed by setting the `colors` property in ansi-to-html constructor to set different colors for each of the 255 ANSI colors when highlighted - but I think this may increase the scope of this feature a bit. Let me know if you think this should be addressed now or later.

amir20:
@Coteh what's the update here? When do you think it will be done?

Coteh:
I believe it's in a good state now. Please feel free to review it. :)

amir20:
I have merged this. Try out the master build. I'll wait a little while before releasing.

Coteh:
Played around with the current master build on my Mac and my Surface. So far no issues from my end, I'll most likely be using this at work over the next few weeks and if I find anything then, I'll create a new issue. Are you going to close #1672 or keep it around for a bit?

gniqizeuy:
keep text highlight css after jumping to context, think it will be useful

PR #1704, opened 2022-04-02
Improve setup documentation

c-w:
Hi there :wave: Thanks for open sourcing this project! I was playing with the code and found a few minor things in the documentation that could have been a little clearer to me as a Go novice:

* Add links to the installation documentation for the various tools the project depends on.
* Make the CLI command to install reflex copy-pastable.
* I also ran into a pnpm user error which dropped a log file to disk which shouldn't be checked in so I added it to the git-ignore list to clean up the git status.

Hope that these minor changes can make it even easier for others to get started to contribute to the repository and thanks again!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 static
 dozzle
 coverage
+.pnpm-debug.log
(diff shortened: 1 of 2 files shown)


amir20:
Thanks!

PR #1705, opened 2022-04-02
Add option to wait for docker before starting

c-w:
## What does this pull request do?

This pull request adds a new configuration option `DOZZLE_WAIT_FOR_DOCKER` which if set will defer the server start until the connection to docker is healthy.

## Why is this useful?

In some environments, the docker connection isn't via a socket file but instead via tcp, for example if we're connecting to a remote docker host or if we're using a proxy to only allow selective access to docker APIs for security reasons. In these scenarios, the docker host may or may not be ready to serve requests at the time that dozzle starts. To simplify deployments in such an environment, it's very useful if dozzle can "self heal" and only start talking to the docker host once the docker host is ready for connections.

Of course it would also be possible to perform an external health check before starting the dozzle container, but this adds non-trivial complexity to the orchestration especially as the the dozzle container doesn't contain a shell so we couldn't for example override the entrypoint to a loop like `sh -c "while ! exec /dozzle; do sleep 5; done"`. Other tools that talk to the docker host (e.g. [Traefik](https://doc.traefik.io/traefik/)) implement this sort of "self heal" mechanism so I believe it would be neat if dozzle could also do this, especially as the added code complexity is quite minor.

## How can I test these changes?

Create a docker-compose file that defines dozzle and the docker socket proxy:

```yaml
version: "2.3"

services:

  dozzle:
    build: .
    environment:
      DOZZLE_NO_ANALYTICS: "1"
      DOZZLE_WAIT_FOR_DOCKER_SECONDS: "120"
      DOCKER_HOST: "tcp://docker-proxy:2375"
    ports:
      - "8080:8080"

  docker-proxy:
    image: tecnativa/docker-socket-proxy
    environment:
      CONTAINERS: "1"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
```

Now start dozzle via `docker-compose up --build dozzle` and you will notice that the server waits to start up since the docker connection can't be established. Now start the docker socket proxy via `docker-compose up -d docker-proxy` and you will notice that dozzle now starts the server.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,6 +33,7 @@ type args struct {
 	Username      string              `arg:"env:DOZZLE_USERNAME" help:"sets the username for auth."`
 	Password      string              `arg:"env:DOZZLE_PASSWORD" help:"sets password for auth"`
 	NoAnalytics   bool                `arg:"--no-analytics,env:DOZZLE_NO_ANALYTICS" help:"disables anonymous analytics"`
+	WaitForDocker bool                `arg:"--wait-for-docker,env:DOZZLE_WAIT_FOR_DOCKER" help:"wait for docker to be available before starting the server."`
 	FilterStrings []string            `arg:"env:DOZZLE_FILTER,--filter,separate" help:"filters docker containers using Docker syntax."`
 	Filter        map[string][]string `arg:"-"`
 }
@@ -46,6 +47,7 @@ var content embed.FS
 
 func main() {
 	var args args
+	var err error
 	parser := arg.MustParse(&args)
 	args.Filter = make(map[string][]string)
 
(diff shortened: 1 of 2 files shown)


c-w:
Note that I had to push a change to the `dev.yml` Github Action as it was failing for me since pull requests from forks don't have access to the secrets. Instead of failing the action, we now skip it for any pull request from a fork.

amir20:
Interesting use case. I never had it happen. But it makes sense. I'll review later.

main.go
```diff
@@ -69,10 +71,16 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("Could not connect to Docker Engine: %v", err)
+	for i := 1; ; i++ {
+		_, err := dockerClient.ListContainers()
+		if err == nil {
+			break
+		} else if !args.WaitForDocker {
+			log.Fatalf("Could not connect to Docker Engine: %v", err)
+		} else {
+			log.Infof("Waiting for Docker Engine (attempt %d): %s", i, err)
```
amir20, line 81: Am I understanding this correctly that it can wait indefinitely? I don't think it should block forever. Common practice is to give up after some time or backoff using something like https://github.com/jpillora/backoff. 

What do you think?

c-w: That works for me. I made a simple change to cap the total amount of waiting time in 9fc3936. Let me know if this works or if we want to pull in a third party dependency to have a fancier retry with more configuration options like polling interval, exponential back-off, etc. I didn't add all of these for now because I didn't want to pollute the settings with too many options until they actually become necessary (for my specific use-case, the simple timeout will be sufficient).

amir20: I like this. I do prefer simple implementation over third party too.

PR #1719, opened 2022-04-19
Cleans up search by removing for loop

amir20:


amir20:
@Coteh adding you to this PR since I was changing your code. I removed the for loop that integrated over all logs as it was causing performance problems. 

I also did some clean up and using hash in the url to trigger the color. I hope in the future to make it work through page navigations too.

Coteh:
I was just about to write a comment when you merged the branch. Very nice solution! Wish I had done it that way from the start.

One thing I noticed after starting it just now using logs from work is that the log lines don't seem to wrap anymore if they're too long. You can easily see this using the split pane view. Once I'm done work I'll try to take some screenshots on my personal machine using dummy logs, if you like.

amir20:
> One thing I noticed after starting it just now using logs from work is that the log lines don't seem to wrap anymore if they're too long. You can easily see this using the split pane view. Once I'm done work I'll try to take some screenshots on my personal machine using dummy logs, if you like.

That's a bug. I just added a new setting to not wrap. But looks like it defaults to not wrapping.

amir20:
Should be fixed now in 436e93c.

PR #1733, opened 2022-05-04
Modifies logs to show only when they are visible

amir20:


amir20:
I won't merge this as it doesn't work well.

PR #1759, opened 2022-06-01
Support for JSON logs

amir20:
This pull request adds JSON support. 

- [x] Show all fields when no fields are selected
- [x] Allow adding and removing json fields
- [x] Save field list to localstorage 
- [x] Support search when JSON is enabled
- [x] Highlight text when searching
- [ ] Filter with json keys

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ static
 dozzle
 coverage
 .pnpm-debug.log
+.vscode
--- a/assets/App.vue
+++ b/assets/App.vue
@@ -107,7 +107,7 @@ function showFuzzySearch() {
     active: true,
   });
 }
-function onResized(e) {
+function onResized(e: any) {
   if (e.length == 2) {
     menuWidth.value = e[0].size;
   }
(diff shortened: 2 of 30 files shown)


VanoDevium:
`JSON.parse()` is very memory hungry for polymorphic objects. 
If there are a lot of them, there may be memory leaks.
Tell me, please, will these JSON-events have a standard structure or a non-permanent one?

VanoDevium:
First, you make a cool tool! I like it! Thank you!

Secondly, what kind of user experience will parsing these logs bring? As an example, key search? Or highlight? Or just pretty screen output? 

> about the JSON.parse(): I also have experience using it with a large flow of messages, it is very slow...

amir20:
Right now this PR brings the ability to show only specific JSON fields. See screenshot which is work in progress. 

Screen Shot 2022-06-02 at 10 08 03 AM

Search by key would be a natural follow up. I think that's technically harder.

VanoDevium:
Just ideas, something might work:

1. Enable compression (gzip) for SSE and return already parsed and pretty formatted JSON 
2. Show as usual text as default view, but when user click on logline - reformat code using JSON.parse()

> Or did I misunderstand you at all?

VanoDevium:
Awesome! 
Sounds good.

Also, this feature can be enabled from startup configuration by the way :)

`Do parse JSON` OR `Do not parse JSON`

amir20:
Fixed some bugs and scrollback works now.

Issue #1766, opened 2022-06-02
DOZZLE_KEY


VanoDevium:
Why not use an irreversible hash of the username and password as the session store key?

Something like:

```
Sha256(DOZZLE_USERNAME + ":" + DOZZLE_PASSWORD)
```

amir20:
That's an interesting idea. And it would make it easier to setup auth. 

Speaking from a security point of view, it should be possible to change the key without updating username or password. Perhaps someone wants to reset all sessions.  I think that use cases is the only one that wouldn't work. 

I don't think it's important though. If you'd like to see this change then a PR is welcomed. Should be easy and the README file needs to be updated.

VanoDevium:
@amir20 
I have very little golang development experience, but I'll try :)

amir20:
I saw you sending another PR. So maybe you are better than you think.

PR #1767, opened 2022-06-02
Auto session storage key

VanoDevium:
#1766

### DIFF
--- a/web/auth.go
+++ b/web/auth.go
@@ -1,6 +1,8 @@
 package web
 
 import (
+	"crypto/sha256"
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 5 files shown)


web/auth.go
```diff
@@ -115,3 +117,8 @@ func (h *handler) clearSession(w http.ResponseWriter, r *http.Request) {
 
 	http.Redirect(w, r, h.config.Base, http.StatusTemporaryRedirect)
 }
+
+func getSessionStorageKey(username string, password string) []byte {
```
amir20, line 121: nit: Call this `generateSessionStorateKey`

amir20 (changes requested):
Nicely done. Just one quick change. I can merge after.

web/auth.go
VanoDevium: done :)

VanoDevium: dmn, you misstype))

PR #1769, opened 2022-06-03
Use vue-json-pretty to display JSON log entries formatted

m-raab:


amir20:
Hey @m-raab, thanks for the PR. It turns out I am already working on JSON support. You can see the PR at https://github.com/amir20/dozzle/pull/1759. 

It would be great if you can contribute to that branch instead. 

A couple notes: 

1. I do want to support filtering the JSON. My containers emit JSON logs with tons of keys. It would be better to limit what the user wants to see. 
2. I like the idea of using `vue-json-pretty`. Maybe we can try it on that branch
3. I parse the JSON on server for faster processing and use JSON for everything now. Make it easier to on the client. 


Thanks again!

PR #1781, opened 2022-06-21
Support building in a container with a custom Dockerfile image.

yavuzsonmez:
"pnpm dev" generates a nodejs error in a container because of the "--open" option of vite which opens the app in the default browser. "pnpm dev:docker" solves this issue and support building dozzle in a container from a custom Dockerfile image.
The "--host" option publishes the default port where the app is running (3000), but it can be changed to another one by adding the preferred port in the script, like: "vite --host 8080".

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,4 +169,4 @@ To Build and test locally:
 2. Install [Go](https://go.dev/doc/install).
 3. Install [reflex](https://github.com/cespare/reflex) with `go get -u github.com/cespare/reflex` outside of dozzle.
 4. Install node modules `pnpm install`.
-5. Do `pnpm dev`
+5. Do `pnpm dev` (or `pnpm dev:docker` to build in a container).
--- a/package.json
+++ b/package.json
@@ -14,8 +14,10 @@
   "author": "Amir Raminfar <findamir@gmail.com>",
   "scripts": {
     "watch:assets": "vite --open",
+    "watch:assets:docker": "vite --host",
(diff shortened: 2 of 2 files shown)


amir20:
I don't understand this error. Why are you getting error? Are you trying to run `dev` inside a docker image?  I don't support that.

yavuzsonmez:
I tried to use "pnpm dev" as entrypoint when a container is instantiated from a custom image. But because the flag "--open" of vite it crashes and and i got a nodejs error:

```
node:events:505
      throw er; // Unhandled 'error' event
      ^

Error: spawn xdg-open ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (node:internal/child_process:289:12)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn xdg-open',
  path: 'xdg-open',
  spawnargs: [ 'http://localhost:3000/' ]
}
 ELIFECYCLE  Command failed with exit code 1.
```

Because "--open" is supposed to open the default browser and there is obviously no browser in the container.

amir20:
Oh so you are trying to use it inside a custom docker image. I don't think that's a popular use case to support. Since it's specific to you, just use `pnpm vite`.

Issue #1810, opened 2022-07-17
Possibility to add a healtcheck


MetalArend:
**Is your feature request related to a problem? Please describe.**
I'm trying to reliably start up a Dozzle container, and check if the service is running as expected. If I do that with for example a DOZZLE_FILTER environment variable with a filter that is not acceptable, the container will be starting, will become running, but will then fail a second (or so) later. It makes checking for the container unreliable right after it starts. It quickly reports to be running, only to still fail right after. But adding sleep 5 in a script feels like an icky way to go about this. 

As has been mentioned in the Dozzle issuetracker before, Docker has a solution to this: adding a healthcheck. I have seen the reasoning to not add the HEALTHCHECK directly in the Dockerfile, and the choice to keep away from things that also have to be maintained without being able to test them. So I wanted to check if I can add something myself. Turns out you are doing the great thing of using multi-stage builds (awesome!), which sadly also means there is no means of in any way adding my own basic healthcheck to the default image, because there's no other possible entrypoint in there than the /dozzle binary. I could recreate my own container, and copy the /dozzle binary and these few other files to that container once again, and add a curl, but I'm never happy to extend good containers, because it uses internal implementation details that might be changing (for good reasons) at any time. So before I go that route, could I pick your brain to resolve it in the Dozzle codebase itself?

**Describe the solution you'd like AND alternatives you've considered**
Would it be possible to have the /dozzle binary have some other endpoints or flags?
- `/dozzle --healthcheck "green"`: check if the Dozzle health is green
- `/dozzle --status "green"`: check if the Dozzle status reports as green
- `/dozzle --stats "containers,filter" --stats-status "containers<10"`: check if the Dozzle reports less than 10 containers to be running 

WDYT? Is there maybe a very easy solution to this in the Dozzle codebase? Some entrypoint that is already being used in the tests, that could be easily promoted to production ready? Or some entrypoint that would make the tests easier as well? And would you consider spending time on it? ;)

amir20:
Hi @MetalArend,

Overall, I am +1 for adding healthcheck. So all good there. Let's figure out a use case first. 

So what is your use case? As far as I know, health checks only work in swarm mode. But I don't support swarm mode completely. How are you using Dozzle and what is your setup? 

> I'm trying to reliably start up a Dozzle container, and check if the service is running as expected. If I do that with for example a DOZZLE_FILTER environment variable with a filter that is not acceptable, the container will be starting, will become running, but will then fail a second (or so) later

[...]

PR #1814, opened 2022-07-19
Adds a simple healthcheck every 2s

amir20:


amir20:
Fixes #1810

MetalArend:
LGTM, only maybe missing some documentation :speak_no_evil:

amir20:
Merged for now. Let's have it in master for a while before adding documentation.

amir20:
@MetalArend See #1819. Sounds like there is a spike on CPU. I guess now it makes sense to just disable healthcheck.

MetalArend:
Shame. I was already looking into how to enable the healthcheck from only the `--health-cmd` flags on docker run, and ran into this: https://github.com/moby/moby/issues/43843

MetalArend:
Basically, it seems impossible to add the same healthcheck as you would do in the Dockerfile with only the docker run flags. I have been trying to set the SHELL in the Dockerfile, have been checking the ENTRYPOINT, but with no luck. I'm hoping anyone from moby itself might have the answer on how to do this, or how to fix it.

amir20:
Well that really sucks. Maybe a short term solution would be to keep the `healthcheck` command, but only enable it with an `ENV` variable. If the variable is off, then it would be a no op. I would have to see if that would still cause a CPU spike. 

On the bug you created, how does using `docker-compose` work? Does it produce the same bug?

amir20:
@MetalArend I think using docker compose works. See solution below...

```
version: "3"
services:
  dozzle:
    container_name: dozzle
    image: amir20/dozzle:latest
    volumes:
    ports:
      - 8080:8080
    environment:
      DOZZLE_LEVEL: trace
    healthcheck:
      test: [ "CMD", "/dozzle", "healthcheck" ]
      interval: 3s
      timeout: 30s
      retries: 5
      start_period: 30s
```

I am able to see `dozzle  | level=trace msg="Executing healthcheck request"`.

So I think this works.

MetalArend:
Yeah, indeed, docker-compose has no issues. But it would be a shame to have to create a yaml file only to run one container :) I would be okay with having to write some strange json to pass it to the health-cmd flag, but it will not accept json :(

Let's see if maybe there is a known work-around. At least one can ask nicely, and hope there is a solution.

corhere:
While you can't (currently) configure a shell-less health check test from the docker CLI, you can use the `--no-healthcheck` flag to disable any check set in the Dockerfile.

I would advise against setting a short health-check interval as each check is rather expensive: https://github.com/moby/moby/issues/33933

amir20:
> I would advise against setting a short health-check interval as each check is rather expensive

Yea I learned that. :) 

The problem with healthcheck is that the initial check will be 0s then 30s. 0s will fail because container is still starting up. 30s later the container will be green status. I wish there was a way to say don't check for another 3s. 

So my (crappy) compromise was to make it check often to unblock use cases where one needs to wait for Dozzle to be green.

corhere:
> The problem with healthcheck is that the initial check will be 0s then 30s. 0s will fail because container is still starting up. 30s later the container will be green status. I wish there was a way to say don't check for another 3s.

We're working on it! https://github.com/moby/moby/pull/40894

PR #1837, opened 2022-08-04
Adds a heartbeat for log stream

### LINKED ISSUE #1831: Cloudflare causing duplicate lines
**Describe the bug**
Related to #1404 but more specific to the underlying issue on Dozzle's side.

The log stream (`stream?id=...&lastEventId=`) is failing each 100,000 milliseconds for unknown reasons. The end result is that the last line is duplicated once the stream reconnects... every 100,000 milliseconds.

This is because the request is automatically resent by the client with the `last-event-id` header. In response, the new stream immediately sends back the log line matching the `last-event-id`, causing that line to be duplicated.

Regardless of what's causing the stream to fail, it should not cause duplicate lines upon a successful reconnection.

**To Reproduce**
Environment:
- Cloudflare (proxy enabled)
- Nginx (reverse proxying to dozzle)

1. Load a container's logs
2. Open inspect element (network tab)
3. Observe failing request each `100000 ms` on the timeline
4. Observe the new stream's payload and confirm it has resent the last (duplicate) line

**Expected behavior**
The line should never be duplicated, either by:
- Changing the client to ignore the first event from a stream if it matches `last-event-id`, or
- Changing the stream to only send events **after** the `last-event-id` (basically change `>=` to `>`)

**Screenshots**
Duplicate log entries:
image

Failing requests (interestingly the first `stream` is working fine, the others fail at exactly 100000 ms intervals):
image

Example retried request with `last-event-id` header:
image

Payload in same request as above showing the duplicate line:
image

**Desktop (please complete the following information):**
 - OS: `Windows`
 - Docker version: `20.10.16`
 - Browser & version: `Chrome 103`
 - Version: `3.12.14`

amir20:
Adds a simple heartbeat to fix timeout errors. Fixes #1831.

### DIFF
--- a/web/logs.go
+++ b/web/logs.go
@@ -105,6 +105,21 @@ func (h *handler) streamLogs(w http.ResponseWriter, r *http.Request) {
 	}
 	defer reader.Close()
 
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	go func() {
+		for {
+			select {
+			case <-r.Context().Done():
+				return
+			case <-ticker.C:
+				fmt.Fprintf(w, "event: container-heartbeat\ndata: \n\n")
+				f.Flush()
+			}
+		}
+	}()
+
 	buffered := bufio.NewReader(reader)
 	var readerError error
 	var message string


amir20:
Tested with `proxy_read_timeout      10s;` and it does work. 

This can be tested with `amir20/dozzle:pr-1837`

wildyaboku:
Can confirm it's working behind Cloudflare as well! 😄

amir20:
Ok merging this for now.

PR #1838, opened 2022-08-04
Tries to add 1 mircosecond to skip first log event

amir20:


amir20:
Great. Thanks for testing. Then I'll get ready to release.

Issue #1867, opened 2022-09-06
Multilingual translation support


AmadeusGraves:
The translation in several languages of this application is needed due to the fact that the other languages are being ignored so that the people who use or try to use this application will have the problem that it is not translated in their language which is unproductive for the application itself since those users will not use this.

AmadeusGraves:
I apologize if I have misspelled or misunderstood, English is not my first language, I also put this task so that when the internationalization is done, help with the translation of this application.

AmadeusGraves:
I can translate spanish

amir20:
@AmadeusGraves That is great feedback. I did some research. I can extract all english messages to configuration files and use Vue's [i18n module](https://vue-i18n.intlify.dev/). Then I could use your help to translate to Spanish. 

I have to make sure everything works before sending it your way.  Does that sound ok?  Would you be able to help with that translations then?

AmadeusGraves:
Of course, it sounds perfect, I'll keep an eye out for when you upload the internationalization and I'll start translating into Spanish.

amir20:
@AmadeusGraves I have created an initial example of _most_ my translations.  You can find `es.yml` at https://github.com/amir20/dozzle/tree/master/locales. 

Can you try translating and send a PR? 

We can test it and see what else is missing. I don't think dates will be updated correctly.

AmadeusGraves:
Perfect, i'll start translating and i help with portuguese, i know a little about this language

amir20:
That would help a lot! 

I am not sure about the dates and numbers. So help me there if you can.

AmadeusGraves:
I have created a pull request, you already have the Spanish and Portuguese translations. Cheers

AmadeusGraves:
Strange, it seems that the dockerhub user and password have not been specified in the tests.
![image](https://user-images.githubusercontent.com/18572939/189460089-74473e8e-681d-46ad-9f85-1a9872987945.png)

amir20:
That's right. I think I added secrets and GitHub doesn't give access to forks. I need to fix it later.

amir20:
Thanks for the PRs! Will check em out.

amir20:
@AmadeusGraves Thanks for the PR. I have merged it for now. You can try `amir20/dozzle:master` to test it. 

Currently, Dozzle automatically picks the locale based on `navigiation.locale`. This might seem confusing to some, so I plan to have a drop down in settings. 

[...]

PR #1870, opened 2022-09-08
Adds i18n support with vue-i18n

amir20:
- Adds asset changes for i18n
- Adds asset changes for i18n
- Adds vite configs


amir20:
Related to #1867

Issue #1835, opened 2022-08-04
Timestamps


JoyZzzzz:
Is it possible to make a setting so that when "show timestamps" is enabled, indicate the exact date "2202.08.21 at" instead of "today at", "yesterday at", "last Tuesday at" and so on?

amir20:
It is possible. :) But I am not sure if it makes sense to do. 

I try to keep Dozzle fairly simple and having lots of settings is not a way to do that. 

Dozzle's main goal is to assist with debugging. Debugging is usually last few days of data which is why I had built it like that initially.

JoyZzzzz:
The days of the week are not convenient for everyone, especially if you don't follow them.. ;)
But the date is always in front of your eyes.

amir20:
I forgot to respond. 

Fair enough. However, I don't want to support multiple formats. Just one that works for "most". 

That said, I haven't seen anybody else request this. I'll keep an eye for other wanting this functionality.  So far most people are ok with the current format and I haven't seen anyone else.

amir20:
@JoyZzzzz After thinking about this a little more, I think it makes sense to completely remove `today`. I came to this conclusion after trying to update locale for different languages. So I will repoen.

AmadeusGraves:
mmmm its posible that this may help. I search for localization in JavaScript and i think that this will be helpfull. https://www.w3schools.com/jsref/jsref_tolocalestring.asp

amir20:
I created a PR for now to use `new Intl.DateTimeFormat`. Let's see if that fixes it.

amir20:
This should be fixed now.

PR #1874, opened 2022-09-12
Uses locale date and time depending on locale

amir20:


amir20:
Fixes time formats. Related to #1867 and #1835

PR #1894, opened 2022-10-01
Fix json log line still says "expand json" when expanded

Coteh:
Noticed this issue when using the new json log lines feature. This should fix it. Now when you expand the json, it will say "collapse json" if it's still expanded.

Screen Shot 2022-10-01 at 5 19 12 PM

Screen Shot 2022-10-01 at 5 19 29 PM

### DIFF
--- a/assets/components/LogViewer/ComplexLogItem.vue
+++ b/assets/components/LogViewer/ComplexLogItem.vue
@@ -4,7 +4,7 @@
       <log-date :date="logEntry.date"></log-date>
     </div>
     <div class="column">
-      <ul class="fields" @click="expanded = !expanded">
+      <ul class="fields" :class="{ expanded }" @click="expanded = !expanded">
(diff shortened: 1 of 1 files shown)


Coteh:
Hey @amir20 just curious if you're going to opt-in this repo for hacktoberfest? If not, then no sweat! Just figured I'd ask.

amir20:
I have no idea what that is. I saw some emails but been so busy with work and kids. 

What is it?

amir20:
I did a quick read. Looks like I would have to create bugs as suggestions for people to create PR?

Coteh:
Yeah it's basically a month long event to encourage new contributors to contribute to open source. You don't necessarily have to create bugs, but if there's any low hanging fruit work you don't have time for, could be a good way to get more contributors contributing.

Hey, no sweat if it's going to be too much for you. I don't think I'll have much time for it this year either.

PR #1896, opened 2022-10-03
Add trending CPU and Memory usage

amir20:
- Refactors to stat history
- Uses markRaw


amir20:
Here is what this looks like right now

Screen Shot 2022-10-11 at 3 16 09 PM

PR #1914, opened 2022-10-18
Adds on mouse over effect for stats

amir20:


EDIflyer:
I'm not sure what this change adds? If I mouseover I can't read the numbers now and although there's a pointer it doesn't seem to do anything?

Mouse not over:
image

Mouseover:
image

amir20:
@EDIflyer I see you have light theme. I didn't really test it well with that theme. 

I need to rethink my colors. It should look like this

Screen Shot 2022-10-20 at 12 40 07 PM

However, it does look like your mouse over is working. It should show the mem/cpu value on hover. With your screenshot, it doesn't look like memory is moving much so maybe it looks like it isn't working when it really just is. 

I'll fix the colors soon and hopefully that will solve your problem.

EDIflyer:
Thanks @amir20 - agree different colours should do the trick on light theme. It doesn't seem to change the data when hovering but will see how the next version is 🙂

amir20:
Flipping the primary and secondary colors looks like this

Screen Shot 2022-10-20 at 2 24 39 PM

Screen Shot 2022-10-20 at 2 24 19 PM

Not sure if I like it

PR #1930, opened 2022-10-29
Fix read username and password from file

dima-bzz:


amir20:
What about the readme file?

amir20:
@dima-bzz the author of go-arg responded. See https://github.com/alexflint/go-arg/issues/202#issuecomment-1295881062

That might be clear implementation than what you did here

main.go
```diff
@@ -99,22 +98,21 @@ func main() {
 	username := args.Username
 	password := args.Password
 	
-	if args.UsernameFILE != "" && args.PasswordFILE != "" {
-		contentUser, err := ioutil.ReadFile(args.UsernameFILE)
+	if args.UsernameFile != "" && args.PasswordFile != "" {
+		contentUser, err := os.ReadFile(args.UsernameFile)
```
amir20, line 102: See if you can use https://github.com/alexflint/go-arg/issues/202#issuecomment-1295881062

amir20 (changes requested):

main.go
dima-bzz: @amir20 How to handle user and password differently?

dima-bzz: My knowledge of go is minimal.

amir20: Not sure I understand. In that comment, it basically does what you did but `UnmarshalText` does reading from file. A little bit cleaner.

dima-bzz: I agree. But I still don't understand how to do it(

amir20: Oh I see. I won't be able to get on laptop until Monday. But basically you want to do exactly what you did in pr. But read the file from UnmarshalText instead. That way you don't read the file in main().

dima-bzz: I almost got it. But I'm getting this error: You can't use args.Password (variable of type *string) as a type string in a struct literal  

```
if args.Username == nil && args.UsernameFile != nil {
        args.Username = &args.UsernameFile.Value
}
```

```
config := web.Config{
                Addr:     args.Addr,
                Base:     args.Base,
                Version:  version,
                TailSize: args.TailSize,
                Username: args.Username,
                Password: args.Password,
        }
```

amir20:
Nice. Looks like you figured it out. Will merge later.

main.go
```diff
@@ -25,15 +24,25 @@ var (
 	version = "head"
 )
 
+type DockerSecret struct {
+  Value string
+}
+
+func (s *DockerSecret) UnmarshalText(b []byte) error {
+  v, err := os.ReadFile(string(b))
```
amir20, line 32: Would it be cleaner if you trim and remove spaces here? I think all Docker secrets would want that.

main.go
```diff
@@ -25,15 +24,25 @@ var (
...
 type args struct {
 	Addr                 string              `arg:"env:DOZZLE_ADDR" default:":8080" help:"sets host:port to bind for server. This is rarely needed inside a docker container."`
 	Base                 string              `arg:"env:DOZZLE_BASE" default:"/" help:"sets the base for http router."`
 	Level                string              `arg:"env:DOZZLE_LEVEL" default:"info" help:"set Dozzle log level. Use debug for more logging."`
 	TailSize             int                 `arg:"env:DOZZLE_TAILSIZE" default:"300" help:"update the initial tail size when fetching logs."`
-	Username             string              `arg:"env:DOZZLE_USERNAME" help:"sets the username for auth."`
-	Password             string              `arg:"env:DOZZLE_PASSWORD" help:"sets password for auth"`
-	UsernameFILE         string              `arg:"env:DOZZLE_USERNAME_FILE" help:"sets the secret path read username for auth."`
-	PasswordFILE         string              `arg:"env:DOZZLE_PASSWORD_FILE" help:"sets the secret path read password for auth"`
+	Username             *string             `arg:"env:DOZZLE_USERNAME" help:"sets the username for auth."`
```
amir20, line 42: Why does this need to be a pointer now? Keep `string` and not `*string` will be cleaner because you can check for `!= ""`

main.go
```diff
@@ -95,28 +106,24 @@ func main() {
...
+	if args.Password == nil && args.PasswordFile != nil {
+		args.Password = &args.PasswordFile.Value
+	}
+
+	if args.Username != nil || args.Password != nil {
+		errorMsg := "Username AND password are required for authentication"
+		if args.Username == nil || args.Password == nil {
+			log.Fatalf(errorMsg)
+		}
+		username = strings.TrimSpace(*args.Username)
```
amir20, line 123: If you see my comments above, you shouldn't need this section anymore

locales/ru.yml
```diff
@@ -0,0 +1,49 @@
+toolbar:
```
amir20, line 1: If you can move this to it's own PR, then that would be great.

amir20 (changes requested):

amir20:
Almost there. Did another round of review and I think your code could be better.

main.go
dima-bzz: If there is a space in the password, it will be deleted.

amir20: Are you sure? https://yourbasic.org/golang/trim-whitespace-from-string/ only spaces beginning and end would be removed. Not in the middle.

dima-bzz: Well, once again, if a space in the password is at the beginning or at the end, then it will be deleted

amir20: I think that's ok, no? If you don't think so then we should just not strip any spaces. I don't know enough about secrets to know what Docker does.

amir20: I know how that works. What I am not sure about different operating system.  For example, in docker doc it says `printf "This is a secret" | docker secret create my_secret_data -` What if they do `"This is a secret "`? 

But I think you are right. No reason to trim it.

main.go
amir20: Then no reason to do this either, right? Let's just make it simple. Revert back `string` without pointer and just check `!=""`. You can remove split and trim here since it's up to the user to make sure that's right.

amir20:
@dima-bzz Quick question. I noticed you are splitting with `\n` anyway for password at https://github.com/amir20/dozzle/blob/master/main.go#L113

What was the reason for this?

dima-bzz:
If you read a file without trimming or splitting a newline, then content read as `test\n`

dima-bzz:
Example  
```
version: "3"
services:
  dozzle:
    build: .
    image: amir20/dozzle_local:latest
    container_name: dozzle
    environment:
        DOZZLE_LEVEL: debug
#        DOZZLE_USERNAME: test
#        DOZZLE_PASSWORD: test
        DOZZLE_USERNAME_FILE: /run/secrets/dozzle_user
        DOZZLE_PASSWORD_FILE: /run/secrets/dozzle_password
    volumes:
    secrets:
        - dozzle_user
        - dozzle_password
    ports:
      - 9191:8080

secrets:
  dozzle_user:
    file: dozzle_user.txt
  dozzle_password:
    file: dozzle_password.txt
```
[dozzle_password.txt](https://github.com/amir20/dozzle/files/9911809/dozzle_password.txt)
[dozzle_user.txt](https://github.com/amir20/dozzle/files/9911810/dozzle_user.txt)

dima-bzz:
@amir20 can be replaced with strings.TrimRight(args.PasswordFile.Value, "\n\r")

amir20:
But I thought that would never happen? I was suggesting we always trim https://github.com/amir20/dozzle/pull/1930#discussion_r1009545648 and you said we don't want that. Am I missing something?

dima-bzz:
We don't want spaces at the beginning or end of the password to be removed. But the newline can be removed.

amir20:
Oh I see. Cool then I can fix that quick. I thought you meant you didn't want any trims.

PR #1933, opened 2022-11-01
Fix removed newline from password file

dima-bzz:


dima-bzz:
@amir20 Well, then we will assign the correctness of the user name to the user. This is about spaces.

amir20:
I don't understand. TrimRight should be used for all values. Is there an example where this would not work?

dima-bzz:
In my opinion, spaces and newline at the beginning and end of the line should be removed in the user name. And only newline should be deleted in the password

amir20:
I don't think I have ever seen a system that cared for spaces in the beginning.  So no reason to over think it. As of right now, I only strip `\r\n` anyway.  Dozzle is a small tool with just me maintaining. I want to keep it simple with less code so it is easier to find bugs. 

Also since Dozzle is a developer tool, hopefully engineers know what they are doing :)

dima-bzz:
Okay, I agree with you. But I would replace trimRight with Trim. I think we can end here.

amir20:
ok. i'll do that.

PR #1936, opened 2022-11-02
Skip Integration Tests and Push branches and PRs from fork

dima-bzz:


amir20:
Amazing. Thanks so much. Tests pass now

dima-bzz:
Check in your repository

amir20:
Looks like it is working

amir20:
By the way I think int test should still run. Just not the docker push part which updates the cache. But now that I see how it can be done I can look at it too.

dima-bzz:
I'm glad to hear that everything worked out

PR #1937, opened 2022-11-02
Skips only the push int tests

amir20:


amir20:
@dima-bzz FYI on this. Moving if skip to the push step. 

Thanks again for finding this solution.

dima-bzz:
@amir20 I was glad to help)

dima-bzz:
@amir20 I'm not sure if integration test is needed because it fails with an error.

image

amir20:
Ah I need to skip that step too.

amir20:
The problem is docker user/pass do not exist

amir20:
See 199c517

dima-bzz:
@amir20 Everything is fine now

image

PR #1938, opened 2022-11-03
Moved logout button

dima-bzz:
I moved logout button to sidebar and add buttons to mobile view. Removed pin a column button from mobile view. But integrations test failed(

### DIFF
--- a/assets/components/SideMenu.vue
+++ b/assets/components/SideMenu.vue
@@ -15,7 +15,7 @@
           </span>
         </button>
       </div>
-      <div class="column is-narrow has-text-right" :class="secured ? px-1 : px-0">
+      <div class="column is-narrow has-text-right" :class="secured ? 'px-1' : 'px-0'">
         <router-link :to="{ name: 'settings' }" active-class="is-active" class="button is-rounded">
           <span class="icon">
             <mdi-light-cog />
--- a/assets/components/SideMenu.vue
+++ b/assets/components/SideMenu.vue
@@ -15,12 +15,19 @@
           </span>
         </button>
       </div>
-      <div class="column is-narrow has-text-right px-0">
+      <div class="column is-narrow has-text-right" :class="secured ? px-1 : px-0">
         <router-link :to="{ name: 'settings' }" active-class="is-active" class="button is-rounded">
           <span class="icon">
             <mdi-light-cog />
           </span>
         </router-link>
+      </div>
+      <div class="column is-narrow has-text-right px-0" v-if="secured">
+        <a :href="`${base}/logout`" active-class="is-active" class="button is-rounded">
+          <span class="icon">
+            <mdi-logout />
+          </span>
+        </a>
       </div>
     </div>
     <p class="menu-label is-hidden-mobile">{{ $t("label.containers") }}</p>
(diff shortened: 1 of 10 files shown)


assets/components/SideMenu.vue
```diff
@@ -15,13 +15,20 @@
...
-      <div class="column is-narrow has-text-right px-0">
+      <div class="column is-narrow has-text-right" :class="secured ? 'px-1' : 'px-0'">
         <router-link :to="{ name: 'settings' }" active-class="is-active" class="button is-rounded">
           <span class="icon">
             <mdi-light-cog />
           </span>
         </router-link>
       </div>
+      <div class="column is-narrow has-text-right px-0" v-if="secured">
+        <a :href="`${base}/logout`" active-class="is-active" class="button is-rounded">
```
amir20, line 26: Add title to this please

assets/pages/Index.vue
```diff
@@ -1,16 +1,5 @@
 <template>
   <div>
-    <section class="hero is-small mt-4">
```
amir20, line 3: OOps. This file shouldn't be there.. Only `index.vue` now. I'll fix it.

locales/en.yml
```diff
@@ -24,7 +24,6 @@ title:
   login: Authentication Required
   settings: Settings
 button:
-  logout: Logout
```
amir20, line 27: Keep these because you need them for tooltip.

amir20 (changes requested):

amir20:
I don't think the UX is that great. Here are some examples

Screen Shot 2022-11-03 at 9 20 23 AM

This looks like it is taking space for no reason. Instead if you want to have settings, why not move it to the expanded menu? Like this where I have highlighted red. 

Screen Shot 2022-11-03 at 9 21 28 AM

I used to have the name of current active container. I must have removed it by accident. 

Now, on desktop I usually have my menu width at a smaller width. So logout cuts off. Hmm I don't know what to do here but I would try to improve it first. 

Screen Shot 2022-11-03 at 9 26 08 AM

Search needs a little padding

Screen Shot 2022-11-03 at 9 32 11 AM


I am really picky about UX :) So of course I am nit picking on some of these. Let's try to clean it up or break this PR to smaller changes so we approve on one change at a time. 

`Index` and `index` shouldn't exist. I'll see if I can fix that.

amir20:
I can fix the int tests. It's probably snapshots that are failing.

amir20:
Rebase. Fixed duplicate files https://github.com/amir20/dozzle/pull/1939/commits/50cc9f2b40fe50bcd499063e0a6c205ae147d201

dima-bzz:
Ok. Close this PR. I create are small PR.

dima-bzz:
There are really problems with the width of the side menu. I wanted to make the width minimal and the buttons not hidden. But min-size is a percentage and, depending on the display, the width is not set correctly.

amir20:
That is true. :( I had the same problem. There is a long history with that menu on the left. It used not be resizable and so many people complained about having long docker labels that it was the only decent option. 

How good are you with UX? Maybe it's time to rethink the menu. At moments like this, I usually cheat and look at other tools and see what they did. :)

dima-bzz:
I have a little experience in UX. I'll think about what I can come up with.

PR #1940, opened 2022-11-03
Fix responsive list containers

dima-bzz:


amir20:
It doesn't look like you pulled the latest changes from master.

amir20:
Done

PR #1942, opened 2022-11-05
Moved logout button to sidebar and fix default width sidebar.

dima-bzz:


amir20:
A lot going on here so give  me sometime while I review it. I have a cold 🤧 so might be slow to respond for a little while.

dima-bzz:
Good. Get well. If you write anything, we will discuss the improvements.

dima-bzz:
1. I made an automatic calculation of the panel width depending on authorization. 
2. Added the saving of the hidden panel in localstorage. 
3. Moved the logout button to the panel and changed its display when hiding the panel. 
4. Removed the margins for logout button in the mobile version. (This is temporary until I take out the buttons in the menu)

amir20:
I am feeling slightly better so giving some feedback. 

I deleted `localstorage` and this is what I get as the default view:

Screen Shot 2022-11-08 at 9 20 47 AM

Still feels tight and not proper spacing.  Also I use a Mac by default but by default on Windows this will cause scroll bars. Which is something to be **avoided** for default view. 

I think it's time to think of doing something _different_. It doesn't all have to be in one line. I did a quick hack moved everything to a new line like so: 
Screen Shot 2022-11-08 at 9 28 06 AM

Also made the icons a little smaller.

Getting this UX right will have big value for people first time using Dozzle. And data shows that I am getting about 200 install per 30 minutes. So I am being very careful of the spacing and visual design. 

Screen Shot 2022-11-08 at 9 31 06 AM


Let me know what you think?

assets/layouts/default.vue
```diff
@@ -74,6 +85,9 @@ function showFuzzySearch() {
     active: true,
   });
 }
+function collapse() {
```
amir20, line 88: I am not seeing why this moved to it's own function.

assets/composables/settings.ts
```diff
@@ -69,7 +72,13 @@ const softWrap = computed({
   set: (value) => (settings.value.softWrap = value),
 });
 
+const collapseNav = computed({
```
amir20, line 75: Ah I see, did you make this a setting? That's cool. Send that in a different PR.

assets/layouts/default.vue
```diff
@@ -112,11 +126,26 @@ function onResized(e: any) {
     left: -40px;
     width: 60px;
     padding-left: 40px;
+    color: white;
```
amir20, line 129: Something doesn't sound right here. Colors shouldn't be hardcoded. Dozzle has light and dark mode. Can you make sure this is the right thing? Doing collapse in light mode looks like this

Screen Shot 2022-11-08 at 9 37 53 AM

That looks broken.

amir20 (changes requested):

assets/layouts/default.vue
amir20: Looks like in master I have bug. The background also turns black. Let's see if we can fix that in this PR. 

Is your intention to have the signout stick out? Why? Isn't the purpose of collapse to hide all buttons. User can uncollapse and click it if needed. No reason to take space.

dima-bzz: If there is no need to show the logout button when the panel is collapsed, then you can remove it and leave it only on the panel.

dima-bzz: I will try to fix the error with the button color.

assets/layouts/default.vue
dima-bzz: There was a compilation error, so I took it out to the function.

assets/layouts/default.vue
amir20: thank you ❤️

dima-bzz:
@amir20 Can you tell me how you reduced the buttons ?

amir20:
@dima-bzz Try adding `is-small` to `.button` class.

dima-bzz:
@amir20 What do you say?

amir20:
I did a quick pull. It looks better. What do you think? Do you like it more?

assets/layouts/default.vue
```diff
@@ -112,7 +114,8 @@ function onResized(e: any) {
     left: -40px;
     width: 60px;
     padding-left: 40px;
-    background: rgba(0, 0, 0, 0.95);
+    color: var(--text-strong-color);
```
amir20, line 117: ❤️

assets/pages/index.vue
```diff
@@ -1,13 +1,9 @@
...
   <div>
-    <section class="hero is-small mt-4">
-      <div class="hero-body">
-        <div class="container">
-          <div class="columns">
-            <div class="column is-narrow" v-if="secured">
-              <a class="button is-primary is-small" :href="`${base}/logout`">{{ $t("button.logout") }}</a>
-            </div>
-          </div>
+    <section class="is-small mt-1" v-if="secured && isMobile">
```
amir20, line 3: You said you will come back to this, right? We can do something similar in mobile and have a row for all the buttons.

amir20 (approved):

amir20:
Files look good. I can merge if you are ok with it.

assets/pages/index.vue
dima-bzz: You can leave the button showing only in the mobile version. But I already have a revision on the buttons for the mobile menu.

dima-bzz:
I don't mind

amir20:
Merging and fixing int tests locally.

amir20:
Something was broken when I merged. Had to fix it with d6e87a7

amir20:
I should have not ignored the int tests. :) https://github.com/amir20/dozzle/actions/runs/3422645762/jobs/5700295282

Issue #1965, opened 2022-11-25
Memory consumption not shown


IrwenXYZ:
**Describe the bug**
Memory consumption is not shown, both in home and individual container pages, but CPU usage is shown correctly. I used the docker compose template provided.

**Expected behavior**
Memory consumption shown correctly in both home and individual container pages

**Desktop (please complete the following information):**
 - OS: Server is Raspberry Pi OS Lite, accessing Dozzle on Windows 11 Pro
 - Docker version
 ```
Client: Docker Engine - Community
 Version:           20.10.21
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:01:19 2022
 OS/Arch:           linux/arm64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.21
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.7
  Git commit:       3056208
  Built:            Tue Oct 25 17:59:41 2022
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.10
  GitCommit:        770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```
 - Browser & version Firefox 107.0, Chrome 107.0.5304.107
 - Version 4.4.1

amir20:
Interesting. There is probably some changes in Raspberry Pi's Docker implementation. I don't have mine handy right now so I can't reproduce. 

Do you know Go? Would you be able to help? I have a possible idea

IrwenXYZ:
Unfortuntely not, I only know some Python. However I can try to look into it if you could point me in the right direction

amir20:
Actually, I think I am wrong. My thought was that https://github.com/amir20/dozzle/blob/master/docker/calculation.go doesn't work correctly. But it actually matches the official Docker CLI https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L239-L249.

So I am out of ideas. Possibly something else is broken. 

I am pretty sick so I won't be able to look at this for a while. Perhaps you can investigate a little by adding logs.

IrwenXYZ:
How do I get the logs? The issue template mentions `--level debug` but I'm not exactly sure where to insert this option.

[...]

PR #1967, opened 2022-11-26
Update readme for memory consumption on ARM devices

IrwenXYZ:
Adds a FAQ point for #1965 where memory consumption wasn't shown on ARM devices.


amir20:
Thanks!

Issue #1970, opened 2022-12-05
Running on multinode swarm


jasunen:
Thanks for this great, simple log viewer!

I have only one issue: **cannot see logs for all the services.**

I have a docker swarm with multiple nodes and starting Dozzle as:

    dozzle:
      image: amir20/dozzle:latest
      volumes:
        - /var/run/docker.sock:/var/run/docker.sock:ro
      ports:
        - "8080:8080/tcp"
      environment:
        DOZZLE_LEVEL: info
        DOZZLE_TAILSIZE: 300
      deploy:
        placement:
          constraints: [node.role == manager]

I am able to see logs for the services running on that **same** node Dozzle starts.
What is the correct way to get logs for all the running services - also on other swarm nodes?

jasunen:
Ah, found this closed issue https://github.com/amir20/dozzle/issues/1608 and understand the challenge. 

Since in my swarm I have only 3 nodes I can use a workaround:
- start dozzle as "global" so each node gets own dozzle service
- do not use service mesh routing - instead publish port in host mode

Of course there are now separated instance of dozzle for each node and I need to connect to UI of all of them separately to see logs for all services.
If amount of nodes would be much higher this would probably not work but as of now it is simple enough for my purposes.

jasunen:
Example of minimal swarm deployment:

    dozzle:
      image: amir20/dozzle:latest
      ports:
        - target: 8080
          published: 8080
          protocol: tcp
          mode: host
      volumes:
        - /var/run/docker.sock:/var/run/docker.sock:ro
      deploy:
        mode: global

If swarm nodes' hostnames are <nodex> (i.e. node1, node2 and node3), logs for each node can be accessed as http://<nodex>:8080

amir20:
Hi @jasunen, thank you for the details in this bug. You are correct that swarm is not well supported. I was hoping to use `mode:global` and some how get all the instances to communicate together. As you saw in the issues, that is a huge under taking which I won't have time. 

Some other users have suggested to have a drop down where the host can be selected.  I think that's a poor experience because personally I think it should just work without knowing what is deployed on each host. 

For those reason, I haven't really spent time on this. I hope maybe some day I can.

amir20:
Closing this

Enissay:
This fix is okay. But then you endup with N instance for N machines which is not ideal but survivable...

[...]

PR #1983, opened 2022-12-12
Adds hostname config for setting UI parameter

amir20:
- Adds hostname config to setting a hostname in title
- Also updates title for hostname


amir20:
Fixes #1970

amir20:
Screenshot 2022-12-12 at 9 34 01 AM

PR #1988, opened 2022-12-14
Typo in README.md

fullbl:
default port is 8080, not 8888!

Also: would be too much to require an env variable for setting the port? I know you can change the external port in docker configuration, but for the other containers it will remain the same!


amir20:
Closing

PR #2006, opened 2023-01-07
Adds sw to vite

amir20:


amir20:
Won't be doing this anymore

Issue #2049, opened 2023-02-16
Separate the stopped containers from the running ones


Enissay:
**Is your feature request related to a problem? Please describe.**
When you have a lot of containers running, many with crashes, showing them together is extremely messy

**Describe the solution you'd like**
It would be great if you separate them in a separate list. You can even make that optional (show together or separate).

**Describe alternatives you've considered**
N/A

**Additional context**
![image](https://user-images.githubusercontent.com/794673/219252323-8145894e-f56b-4c77-9545-af3a7602cec8.png)

amir20:
This is interesting. I guess I just assumed people would clean up their stopped containers. But I think it makes sense. Let's keep this open and hopefully I'll get to it.

amir20:
@Enissay what do you think of `amir20/dozzle:pr-2049`? I group to different sections now. But only if the option to show stopped containers is enabled.

amir20:
See screenshot on #2058 

I am not sure if I like it. My use case is to look at similar containers regardless of running or not.

Enissay:
Good enough...

If you don't like it, can you make the split optional ? so the user can choose between the two ?

Enissay:
Offtopic question, what is the use of the `amir20/echo` image ? I could not tell from the [dockerfile](https://hub.docker.com/layers/amir20/echo/latest/images/sha256-8f8d6f9bc3a32e799fab5ef3db931922a6c1bbabafad48cfdbff7b6651418501?context=explore)... I am curious as it seems useful ^^

amir20:
> Good enough...
> 
> If you don't like it, can you make the split optional ? so the user can choose between the two ?

I am trying to keep all the different options to a minimum 😆. Otherwise, I'd have to support too many combinations. 

Let me think about it. Have you played with it? I personally don't have stopped containers enabled often so maybe it doesn't matter. 

> re: amir20/echo

I use if for my testing. Code is at https://github.com/amir20/echo. I can do 

```
docker run -i amir20/echo < lorem.txt
```

And it will randomly choose logs from `lorem.txt` and send it to `stdout`. When I was doing JSON support I had to check a lot of different outputs so I needed something more robust.

amir20:
It doesn't look like I have the time to spent on this. I think I want to redesign some of the menu on the left but for now I will close this. 

> Have you played with it?

You can use the `pr-2058` for now.

amir20:
I saw another tool just put the running containers on top. I'll do the same.

PR #2058, opened 2023-02-21
Groups running and stopped containers in different menus. See #2049

amir20:
Screenshot 2023-02-21 at 11 25 08 AM


amir20:
Docker desktop seems to be just mix everything. I'll close this for now as I haven't had time to look at it and a bunch of tests will fail if I merge.

PR #2059, opened 2023-02-21
Adds support for multiple hosts

amir20:
This PRs adds the ability to connect to remote hosts via TCP with TLS. SSH is not supported. The new command line introduced is `--remote-host`. ENV vars are also supported via `DOZZLE_REMOTE_HOST`.  The certs will need to be mounted in the `/cert` folder. 

Note that local `docker.sock` is still required. 

Here are examples of running Dozzle: 

```
docker run -v /var/run/docker.sock:/var/run/docker.sock -v /path/to/certs:/certs -p 8080:8080 amir20/dozzle --remote-host tcp://167.99.1.1:2376
```

This will connect to `167.99.1.1` via TLS by looking for appropriate certs in `/certs/{ca,cert,key}.pem` or `/certs/{host}/{ca,cert,key}.pem` so that it can work with multiple hosts. 

To use ENV variables: 

```
docker run -e DOZZLE_REMOTE_HOST=tcp://167.99.1.1:2376,tcp://167.99.1.2:2376 -v /var/run/docker.sock:/var/run/docker.sock -v /path/to/certs:/certs -p 8080:8080 amir20/dozzle
```
This will connect to two hosts using the comma separated values to `DOZZLE_REMOTE_HOST`.


amir20:
Hmm looks like ssh within the container might not work....still investigating

amir20:
https://github.com/amir20/dozzle/#connecting-to-remote-hosts has been updated.

PR #2099, opened 2023-03-28
Adds health status by parsing the label from status

amir20:
Adds healthcheck test by looking status value. 

Screenshot 2023-03-29 at 8 54 21 AM

### DIFF
--- a/assets/components.d.ts
+++ b/assets/components.d.ts
@@ -10,6 +10,10 @@ export {}
 declare module '@vue/runtime-core' {
   export interface GlobalComponents {
     CarbonCaretDown: typeof import('~icons/carbon/caret-down')['default']
+    'Cil:checkAlt': typeof import('~icons/cil/check-alt')['default']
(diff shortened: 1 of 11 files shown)


EDIflyer:
Would it be worth adding something explaining this within the app?  Perhaps in the about section in the settings page there could be a wee explainer? It's just I was looking at Dozzle today and was confused why some containers had green ticks and others didn't - I hovered over the icon but of course it just changed to show the 'pin as column' option. Had a search through the changelogs and came across this, then cross-referenced in Portainer to see it was indeed the ones with status 'healthy' rather than just 'running' that got the ticks!

(PS as ever thanks for such a nice app for viewing log files!!)

amir20:
I was planning to add tooltips. But hmmmm now I realize that wouldn't work haha. Perhaps a better solution would be when container clicked, at the top it would show a similar error icon with a tooltip? 

But yea I see your point.

PR #2119, opened 2023-04-11
ci: use docker bake instead

amir20:


amir20:
I am using https://www.deploysentinel.com/blog/docker-buildx-cache-with-github-actions but it doesn't look caching improved anything.

PR #2145, opened 2023-04-24
feat: makes swarm id lighter

### LINKED ISSUE #2144: Idea: don't show hash in container names
**Is your feature request related to a problem? Please describe.**

When having lots of containers, it can look like word salad, and cluttered or "noisy" trying to look through the list. (I know there's a search but sometimes I just want to quickly scan with my eyes), and make finding the container you want more "eye-fatiguing" than necessary.

**Describe the solution you'd like**

Since hashes aren't often necessary for cases like this, it would be nice to have an option for users who don't choose to see the hashes. Perhaps default that option to show the full hash as it currently does, and then the user can go into the settings and choose to hide them.

Then you'd be left basically with just the name of the container, and if in a swarm, the swarm name and container name, both of which are actual "human readable" words rather than random chars.

**Additional context**

Here's an example screenshot of my list. Look at all that noisy, unhelpful, ugly random hashing in each container. :)

image

As an example, I drew a line through the hashes to show what to hide. IN the case of more than 1 of the same container, an idea may be to only show the 1st letter of the hash to distinguish it (see the second item with a line through it that shows only the first letter of the hash).

amir20:
Fixes #2144

### DIFF
--- a/assets/components/LogViewer/ContainerTitle.vue
+++ b/assets/components/LogViewer/ContainerTitle.vue
@@ -6,7 +6,7 @@
         <cil:x-circle v-else-if="container.health == 'unhealthy'" />
         <cil:circle v-else />
       </span>
-      {{ container.name }}
+      {{ container.name }}<span v-if="container.isSwarm">.{{ container.swarmId }}</span>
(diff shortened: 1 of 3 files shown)


amir20:
Screenshot 2023-04-24 at 11 44 48 AM

PR #2200, opened 2023-05-18
Use ui-monospace as the default font for log viewer

agneevX:
For the most part, this makes Safari use SF Mono instead of falling back to Consolas.


amir20:
Thanks! I'll look at it on my laptop before merging.

PR #2331, opened 2023-07-31
feat: adds bar charts back to dashboard table

amir20:


amir20:
168bd249769038f6ddc45a3804634813

PR #2365, opened 2023-08-24
feat: Improve page title and add localization

codeman1o1:
This PR improves the page title handling.
It fixes the following things:
* Add translations
* Add pluralization for the amount of running containers
* Prevent unnecessary updates by only setting the title when the data is fully loaded
* Set the correct title for the settings page when reloading
* A bug where the title would not change when switching from a container logs page to the dashboard

_Note_
I tried my best with the translations, but I do not know if they are all correct.
Google Translate gives me the right English translation back for each language.

### DIFF
--- a/assets/pages/container/[id].vue
+++ b/assets/pages/container/[id].vue
@@ -16,11 +16,8 @@ const { id } = defineProps<{ id: string }>();
 const currentContainer = store.currentContainer($$(id));
 const { activeContainers, ready } = storeToRefs(store);
 
-setTitle("loading");
-
-onMounted(() => {
-  setTitle(currentContainer.value?.name);
+watchEffect(() => {
+  if (ready.value === true)
+    currentContainer.value !== undefined ? setTitle(currentContainer.value.name) : setTitle("Not Found");
 });
-
-watchEffect(() => setTitle(currentContainer.value?.name));
 </script>
--- a/assets/pages/index.vue
+++ b/assets/pages/index.vue
@@ -52,9 +52,13 @@
 <script lang="ts" setup>
 import { Container } from "@/models/Container";
 
+const { t } = useI18n();
 const { version } = config;
 const containerStore = useContainerStore();
-const { containers } = storeToRefs(containerStore) as { containers: unknown } as { containers: Ref<Container[]> };
+const { containers, ready } = storeToRefs(containerStore) as unknown as {
+  containers: Ref<Container[]>;
+  ready: Ref<boolean>;
+};
 
 const mostRecentContainers = $computed(() => [...containers.value].sort((a, b) => +b.created - +a.created));
 const runningContainers = $computed(() => mostRecentContainers.filter((c) => c.state === "running"));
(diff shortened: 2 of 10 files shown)


assets/pages/container/[id].vue
```diff
@@ -16,11 +16,8 @@ const { id } = defineProps<{ id: string }>();
 const currentContainer = store.currentContainer($$(id));
 const { activeContainers, ready } = storeToRefs(store);
 
-setTitle("loading");
```
Comment (amir20) [inline assets/pages/container/[id].vue:19]: But when loading this is removed now. Perhaps in `watchEffect` it should check for `ready` and set loading depending on that?

assets/pages/index.vue
```diff
@@ -52,9 +52,13 @@
 <script lang="ts" setup>
 import { Container } from "@/models/Container";
 
+const { t } = useI18n();
 const { version } = config;
 const containerStore = useContainerStore();
-const { containers } = storeToRefs(containerStore) as { containers: unknown } as { containers: Ref<Container[]> };
+const { containers, ready } = storeToRefs(containerStore) as unknown as {
+  containers: Ref<Container[]>;
```
amir20, line 59: I always thought it was weird that I had to do this. I didn't understand why `storeToRefs` doesn't just know the type.

assets/pages/container/[id].vue
```diff
@@ -16,11 +16,8 @@ const { id } = defineProps<{ id: string }>();
...
-
-onMounted(() => {
-  setTitle(currentContainer.value?.name);
+watchEffect(() => {
+  if (ready.value === true)
+    currentContainer.value !== undefined ? setTitle(currentContainer.value.name) : setTitle("Not Found");
```
Comment (amir20) [inline assets/pages/container/[id].vue:21]: Maybe here it could just set loading if not ready yet?

Comment (codeman1o1) [reply assets/pages/container/[id].vue:21]: The reason I removed the loading title, is because it is loaded almost instantly.
It only needs to load the most basic container information, and not the logs.
Also, if the title would be set to loading first, it would almost instantly be set to the container name, which doesn't look very good IMO

assets/pages/container/[id].vue
Comment (codeman1o1) [reply assets/pages/container/[id].vue:19]: I think this conversation could be closed in favor of https://github.com/amir20/dozzle/pull/2365/files#r1304933814

assets/pages/container/[id].vue
Comment (amir20) [reply assets/pages/container/[id].vue:21]: > The reason I removed the loading title, is because it is loaded almost instantly.

That might be true for you. :) And I have put a lot of effort to make sure that it is fast. However, there is probably about 2% of Dozzle users who have multiple hosts with 100s of containers. That probably takes about second to load which is why I added the loader screen when there is no data. 

I am still testing so let me see how it looks.

Comment (amir20) [reply assets/pages/container/[id].vue:21]: I tested locally and it does take 2 seconds to load. However, I think the experience is fine. So I'll merge.

amir20:
This is released. Thanks for your contributions!

Issue #2437, opened 2023-10-21
Option of settings being at server/deployment level?


EDIflyer:
**Is your feature request related to a problem? Please describe.**
I use Dozzle (which is totally brilliant BTW!) on multiple different computers and browsers.  Currently I need to set all my customisations anew on each browser I'm using.

**Describe the solution you'd like**
The option for the settings to apply across all browsers - I was thinking this could be particulary useful for remembering upcoming filter settings (#2432), pinned/stored searches (#2036), etc.

**Describe alternatives you've considered**
Continue the current model

**Additional context**
I saw your helpful post in #1859 where you explained the three levels of config in Dozzle:
> There are three scopes of configurations in Dozzle:
> 1. User scope which is specific to a browser (localstorage)
> 2. Deployment scope which is per install and configured for all users
> 3. And no scope, something like hardcoded in binary

I guess what I'm asking for is the option to, on a per-deployment basis, switch to have the settings applying at level 2 rather than level 1.  Although this could be an option in the UI I wonder if it might work easier/better as an environment variable used when running the docker container (e.g., a `--server_settings` flag in docker run or `DOZZLE_SERVER_SETTINGS: true` in docker-compose) - this would get around any issues for setups where there _are_ multiple users and you wouldn't want one user overriding another's settings and obviously wouldn't change the current default of level 1.

I was envisaging this working something along the lines of any setting changes still being made to `localstorage` as normal but then also being sent to the server. When first loading the Dozzle app it would then ask if there were any settings stored on the server and use them to populate `localstorage`.  If you made a setting change on one browser then a manual refresh on the other browsers would be required to reload the settings.

If this is too big/complicated/niche a request don't worry - it just would save me having to remember to change a setting across lots of browsers and as mentioned above I was thinking it might become even more handy going forward if users start to have more by way of customisations.

Hope the above makes sense and thanks again for all your work on this!

amir20:
Thanks! I have been thinking about this a lot. There are two parts: 

Technically, I was thinking just storing everything as json. I also thought about using some kind of embedded database but I think that's too complicated. 

The harder part is configuration. 

[...]

PR #2442, opened 2023-10-23
feat: adds forward-proxy

amir20:
See #2437 on discussion to create multi-user options for Dozzle. 

This is a start for saving all configuration on disk. 

Screenshot 2023-10-23 at 12 56 10 PM

This PR introduces `Remote-*` headers to associate each session with a user. 

```
Remote-Email: admin@example.com
Remote-Groups: admins,dev
Remote-Name: Admin
Remote-User: admin
```

To use this, set `DOZZLE_AUTH_PROVIDER` to `forward-proxy`. Here is an example from compose file I used for testing....

```

  dozzle:
    image: amir20/dozzle:latest    
    networks:
      - net
    environment:
      DOZZLE_AUTH_PROVIDER: forward-proxy
    volumes:
    labels:
      - 'traefik.enable=true'
      - 'traefik.http.routers.dozzle.rule=Host(`dozzle.example.com`)'
      - 'traefik.http.routers.dozzle.entrypoints=https'
      - 'traefik.http.routers.dozzle.tls=true'
      - 'traefik.http.routers.dozzle.tls.options=default'
      - 'traefik.http.routers.dozzle.middlewares=authelia@docker'
    expose:
      - 8080
      
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ COPY --from=node /build/dist ./dist
 COPY analytics ./analytics
 COPY healthcheck ./healthcheck
 COPY docker ./docker
+COPY auth ./auth
 COPY web ./web
 COPY main.go ./
 
--- a/assets/pages/index.vue
+++ b/assets/pages/index.vue
@@ -1,5 +1,13 @@
 <template>
   <div class="flex flex-col gap-16 px-4 pt-8 md:px-8">
+    <section v-if="config.user">
+      <div class="flex items-center gap-4">
+        <div class="ml-auto">
+          {{ config.user?.email }}
+        </div>
+        <img class="h-10 w-10 rounded-full p-1 ring-2 ring-base-content/50" :src="config.user.avatar" />
+      </div>
+    </section>
     <section>
       <div class="stats grid bg-base-lighter shadow">
         <div class="stat">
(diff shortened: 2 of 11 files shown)


amir20:
This can be tested with `amir20/dozzle:pr-2442`.

EDIflyer:
Yep, works well - at first I thought nothing was happening then realised I had to set the environment variable - my fault! 🤦 

Interestingly for Portainer it didn't seem to like the setup above and I had to do it slightly differently
``` yaml
    environment:
      - PATH=/bin
      - TZ=Europe/London
      - DOZZLE_AUTH_PROVIDER=forward-proxy
```
When I tried just the `DOZZLE_AUTH_PROVIDER` with a colon it rejected it saying it had to be a string. Then when I switched all 3 to colons it would deploy but didn't seem to work as the user email/logo vanished.  Perhaps just something I don't quite understand about YAML and/or Portainer and/or Docker Compose!

amir20:
@EDIflyer merging this. Will follow up later this week to store the data. I'll release 5.1 when I update the documentation.

EDIflyer:
Ideal. BTW just twigged the reason for the difference in our docker compose files - I'd put a hyphen at the start of each of my environment variables so that was why I needed to use `=` rather than `:` 👍😉

PR #2445, opened 2023-10-25
feat: syncs settings to disk for authenticated users

amir20:
For authenticated users, settings is now saved to `/data/[username]/settings.json`. This location will also be used for other user stuff like `savedSearch.json`. 

User should be able to do the following: 

1. Go to a browser and make sure there is a valid user signed in 
2. Make some changes to settings 
3. Go to a new browser that has never used Dozzle. Those settings should be copied over from server. 
4. If needed, `--volume /path/to/data:/data` is needed to persist settings between dozzle restart. However, Dozzle also stores content to `localStorage` so it will also sync localStorage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 certs
+data
 dist
 node_modules
 .cache
--- a/assets/auto-imports.d.ts
+++ b/assets/auto-imports.d.ts
@@ -363,7 +363,7 @@ declare global {
 // for type re-export
 declare global {
   // @ts-ignore
-  export type { Component, ComponentPublicInstance, ComputedRef, InjectionKey, PropType, Ref, VNode } from 'vue'
+  export type { Component, ComponentPublicInstance, ComputedRef, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, VNode, WritableComputedRef } from 'vue'
 }
 // for vue template auto import
 import { UnwrapRef } from 'vue'
--- a/assets/auto-imports.d.ts
+++ b/assets/auto-imports.d.ts
@@ -363,7 +363,7 @@ declare global {
 // for type re-export
 declare global {
   // @ts-ignore
-  export type { Component, ComponentPublicInstance, ComputedRef, ExtractDefaultPropTypes, ExtractPropTypes, ExtractPublicPropTypes, InjectionKey, PropType, Ref, VNode, WritableComputedRef } from 'vue'
+  export type { Component, ComponentPublicInstance, ComputedRef, InjectionKey, PropType, Ref, VNode } from 'vue'
 }
 // for vue template auto import
 import { UnwrapRef } from 'vue'
--- a/assets/auto-imports.d.ts
+++ b/assets/auto-imports.d.ts
@@ -12,14 +12,14 @@ declare global {
   const $ref: typeof import('vue/macros')['$ref']
   const $shallowRef: typeof import('vue/macros')['$shallowRef']
   const $toRef: typeof import('vue/macros')['$toRef']
-  const DEFAULT_SETTINGS: typeof import('./composables/settings')['DEFAULT_SETTINGS']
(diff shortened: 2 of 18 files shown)


amir20:
This should be ready for testing @EDIflyer. I think you should be able to just install and validate settings are stored. You can mount `/data` if you want to see the data.

EDIflyer:
OK have tried this and seems to work 🙂 

As instructed I added a volume to my docker compose so I could view the `settings.json` then went to a logged in version and made some settings changes. These did indeed get saved in `/[specified directory]/[username]/settings.json`

I then went to a new browser, logged in with the same credentials on Authelia and it showed the expected settings.

So the only bit I was expecting to work that didn't was a [Ctrl-[F5] on a browser that was already using Dozzle, but presumably that's not enough to reload the settings and would require the stored site data to be cleared.  Perhaps worthy of a UI button to 'load settings from server'??

amir20:
> So the only bit I was expecting to work that didn't was a [Ctrl-[F5] on a browser that was already using Dozzle, but presumably that's not enough to reload the settings and would require the stored site data to be cleared. Perhaps worthy of a UI button to 'load settings from server'??

Not sure what you mean ctrl-F5. But yes, `localStorage` takes priority over server configuration. I thought about this a lot. I didn't know what would make sense.  Right now, server settings are only used when `localStorage` doesn't exist.  I could make it so that server settings always overwrite `localStorage`.  But I don't know if that would make sense in all instances. 

What do you think?

EDIflyer:
Sorry I was just meaning that Ctrl-F5 is a fuller reset on a browser page so normally causes cached items to reload.

As use cases might be different I wonder if the best sequence might be:
- no `localStorage` exists, no server config - load Dozzle defaults
- no `localStorage` exists, server config present - load server config
- `localStorage` exists, no server config - continue as older versions but also save any setting changes to the server
- `localStorage` exists, sever config exists - continue to prefer `localStorage` however in settings offer the user a button to 'load settings from server' which would then overwrite `localStorage`

It's just if the last bit doesn't exist then server settings would only be used for new browsers and the only way to load them for old browsers would be to clear site settings (which is normally a bit of a hassle) - at least this way we'd be giving users who want to sync settings the ability to do so, but without forcing it on them?

amir20:
> localStorage exists, sever config exists - continue to prefer localStorage however in settings offer the user a button to 'load settings from server' which would then overwrite localStorage

What would be the harm to always overwrite? I am just trying to avoid a button that probably won't get used much. Also, I wouldn't know if it's out of sync really. I just know they are different.  What's an instance where the user wouldn't want to always sync when they are signed in?

EDIflyer:
I'd want that personally but thought you had maybe done that on purpose and were preferring `localStorage` by design 🙂

I guess the main reason not to do it is some users may want different settings on different browsers but presumably they would then not just bothering using user ID pass through at all?

amir20:
Yea, I was thinking that too. I can't think of any product behaving that way. If I store some settings for a product then I would expect it to sync across devices. I'll keep thinking about it. I might change my mind when I try to code it up :P

amir20:
Ok try again. I make it always overwrite now. Then went on a different browser and noticed it updated after a refresh. 

Tell me if it feels more natural. I think it feels right for me. At least enough to want to an explicit button.

EDIflyer:
Just tried and I agree that seems much more natural and obvious to me that I change settings in one place and when I hit refresh on another browser they're imported from the server. I think it removes a level of potential confusion around changes not being reflected.

The only very _very_ minor thing that I thought may confuse a user is if you change settings in one place whilst you have Dozzle already running in another browser and you navigate to the settings page and change something - the settings on that page now overwrite the server settings. I don't support it would be possible to warn the user if they go to the settings page and they're out of sync with the server and warn them to refresh the page? (I'm guessing you don't want to force refresh on navigation to the settings page).  As I Say very minor and not a big issue given the fairly small number of settings available to change anyway but just something that occurred to me! (hope my description makes sense!!)

EDIflyer:
Just to confirm tried latest cffe751 build and still works well.

PR #2492, opened 2023-11-12
docs: update authentication.md 

sureshram:
looks like the authentication is case sensitive.

### DIFF
--- a/docs/guide/authentication.md
+++ b/docs/guide/authentication.md
@@ -185,7 +185,7 @@ Dozzle supports multi-user authentication by setting `--auth-provider` to `simpl
 ```yml
 users:
   admin:
-    name: "Admin"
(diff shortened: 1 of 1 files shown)


docs/guide/authentication.md
```diff
@@ -185,7 +185,7 @@ Dozzle supports multi-user authentication by setting `--auth-provider` to `simpl
 ```yml
 users:
   admin:
```
amir20, line 187: @sureshram this line is actually the username.

Line 188 is the display name. Like "John Smith"

sureshram: I see, The confusion is that, it's not clear which one is the username field. "admin" sounds like a role. Added commit make it more clear.

amir20:
🎉

PR #2526, opened 2023-11-21
feat: when doing modifier+f again in search, hides search and show browser search

amir20:


flamestro:
LGTM
- tested by using amir20/dozzle:pr-2526

PR #2528, opened 2023-11-22
fix: update de.yml to use consistent wording and fix small issues

flamestro:
- In german you can use "ihr/Ihr" or "du/Du" to refer to someone. Dozzles translation switches between both which is not a good way to do it. We should stay consistent with one variant.  ihr/Ihr is considered very formal while du/Du is considered more direct. Writing it with an uppercase start character is considered to be the more formal variant in both scenarios. In this commit all ihr/Ihr usages are replaced with "Du".

- some translations like 'running / total containers' were in the wrong form and got fixed. This is because containers refers to the plural so running and total need to refer to the plural as well.

- Some translations did not get fixed as I could not think of a better term. For instance, "AVG MEM" and "AVG CPU" would become to long "Durchschn. Speicher" or "Durchschn. CPU"

### DIFF
--- a/locales/de.yml
+++ b/locales/de.yml
@@ -6,8 +6,8 @@ toolbar:
   show-all: Zeige alle Streams
 label:
   containers: Container
-  total-containers: Gesamt Container
-  running: Laufend
+  total-containers: Gesamte Container
+  running: Laufende
   total-cpu-usage: Gesamte CPU Auslastung
   total-mem-usage: Gesamte Speicher Auslastung
   dozzle-version: Dozzle Version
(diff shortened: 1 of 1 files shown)


amir20:
Thanks! There is a section that I totally forgot to translate. 

https://github.com/amir20/dozzle/blob/b243783f5de092a51a0e4f6e7113cc215e48afb6/locales/en.yml#L84-L90

Would you be able to add this too? :)

flamestro:
Of course, let me take a look

amir20:
Thanks!

Issue #2546, opened 2023-11-29
Build instructions do not build a fresh project.


ab623:
**Describe the bug**
I have been trying to build the project from scratch and come across a build error following the readme steps.

**To Reproduce**
1. Create a brand new directory
2. Clone the project
3. Run the Build steps in the Readme
4. The project returns and error `[backend] [00] main.go:65:12: pattern all:dist: no matching files found`. 

**Expected behaviour**
The project to build

**Analysis + Resolution**
Through a bit of research I found that the go application was trying to embed the `dist` directory, which at the point of a fresh clone does not exist. 

To fix the situation I had to run `pnpm build` which then created the directory and then `pnpm dev` worked as expected. 

Either that step is added into the build steps, or an empty dist folder is created and committed to the library.

PR #2547, opened 2023-11-29
fix: Added .gitkeep to dist folder, to enable fresh build with no errors

ab623:
fix: #2546 

Added a .gitkeep file to the dist folder, so that the dist folder can be committed to the repo.

This will enable new projects to be cloned and built without causing any missing folder issues.


amir20:
Thanks!

Issue #2532, opened 2023-11-25
Adding a *Restart* option for containers


akash-ramaswamy:
**Is your feature request related to a problem? Please describe.**
I felt many times that it would be so nice to have a **Restart** option for individual containers, since it would be handy for immediately restarting the containers during development.

**Describe the solution you'd like**
Adding a refresh button here and mapping this to restart using [docker engine api](https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerRestart)

![image](https://github.com/amir20/dozzle/assets/87078332/1b214da0-b358-41f2-b63f-1b10215fbf56)

I would be happy to get this assigned to me if @amir20 you feel that this is a good handy feature!

akash-ramaswamy:
We can use the [golang SDK](https://pkg.go.dev/github.com/docker/docker/client?utm_source=godoc#section-documentation) for this feature!

amir20:
Hi @akash-ramaswamy. The focus of Dozzle has been so far to collect data like logs and stats. I haven't focused on other uses cases which would mutate or change the state of containers. I think there are other tools like Portainer which do same thing. 

Additionally, taking space in the UI for a button which won't be used by a lot of people seems like a bad decision.

akash-ramaswamy:
Hi @amir20 !

I understand the focus on data collection, but a Restart option for individual containers in Dozzle could significantly boost developer productivity during active development phases. It aligns with common needs and could be used as "optional" setting or so, to address UI concerns. Your thoughts on this?

amir20:
Hey @akash-ramaswamy, I hear you. But I am very careful to add new features, especially if it won't be useful for most users. 

> Restart option for individual containers in Dozzle could significantly boost developer productivity during active development phases

Thinking about this personally, I have never had the need to restart a container. I am usually jumping between my terminal or created some kind of process in Docker Desktop that automates this. 

> It aligns with common needs and could be used as "optional" setting or so, to address UI concerns

There aren't actually a lot of bells and whistles to Dozzle right now. But you also have to consider the maintenance. As Dozzle has gotten more sophisticated in the last year, I have to be extra careful about releasing and not breaking anything.  This is one the of the major reasons why I don't add a lot of hidden features because then I would have to test them all. Some times my integration tests doesn't catch anything. 

[...]

PR #2548, opened 2023-11-29
feat: can start and stop containers now using the drop down. This feature is not enabled by default. 

akash-ramaswamy:
Added actions for containers #2532 

Screenshot from 2023-11-30 02-27-27


internal/web/container_actions.go
```diff
@@ -0,0 +1,28 @@
...
+
+func (h *handler) containerActions(w http.ResponseWriter, r *http.Request) {
+	action := chi.URLParam(r, "action")
+	id := chi.URLParam(r, "id")
+
+	for _, client := range h.clients {
+		cont, err := client.FindContainer(id)
+
+		if err == nil {
+			if client.ContainerActions(action, cont.ID) != nil {
```
amir20, line 17: it would be better to get the err and log it with `log.Error` since users might be confused why it isn't working. 

Also looking at the UI, it looks like you are showing a toast. But you don't return anything here, is that intentional?

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,83 @@
...
+        type: "error",
+        message: message ?? "Something went wrong",
+        title: "Action failed",
+      },
+      { expire: 5000 },
+    );
+
+  actionStates[`${action}-action`] = true;
+
+  await fetch(withBase(actionUrl))
```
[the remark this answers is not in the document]
amir20: As mentioned later, this should also be post. All APIs that change something should be post.

amir20 (changes requested):
Alright, I just provided A LOT of feedback. Hope all makes sense. I won't be able to review again until tomorrow. Good work so far!

internal/web/container_actions.go
```diff
@@ -0,0 +1,28 @@
...
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+)
+
+func (h *handler) containerActions(w http.ResponseWriter, r *http.Request) {
+	action := chi.URLParam(r, "action")
+	id := chi.URLParam(r, "id")
+
+	for _, client := range h.clients {
```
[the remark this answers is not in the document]
amir20: Forgot to mention that you need to pass the host to the URL. Like I do in https://github.com/amir20/dozzle/blob/050e499f8e718b959cf3e26808947203d5eed860/assets/composable/eventsource.ts#L88

internal/web/container_actions.go
akash-ramaswamy: Yes, I have added error logs now.

akash-ramaswamy: Actually i wanted to notify error toast based on the `statusCode` recieved!

assets/components/LogViewer/LogActionsToolbar.vue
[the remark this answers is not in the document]
akash-ramaswamy: > Huh? I think you can just `await` with `try-catch` right?

I wanted to show toast based on `statusCode`! But i have refactored it now.

internal/docker/client.go
```diff
@@ -145,6 +149,23 @@ func (d *Client) FindContainer(id string) (Container, error) {
 	return container, nil
 }
 
+func (d *Client) RestartContainer(id string) error {
+	return d.cli.ContainerRestart(context.Background(), id, container.StopOptions{})
```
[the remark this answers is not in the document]
akash-ramaswamy: Yes, I have removed.

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,83 @@
...
+    showToast(
+      {
+        type: "error",
+        message: message ?? "Something went wrong",
+        title: "Action failed",
+      },
+      { expire: 5000 },
+    );
+
+  actionStates[`${action}-action`] = true;
```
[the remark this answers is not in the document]
akash-ramaswamy: > Actually when I think about it, shouldn't there only be one action state? If container is restarting then both stop and restart should be disabled, right?

This is a valid point, But i couldn't find a better way to do this without three separate states. but i have now moved this to a composable by injecting container!

assets/composable/containerActions.ts
```diff
@@ -0,0 +1,47 @@
+type ContainerActions = "start" | "stop" | "restart";
+export const useContainerActions = () => {
+  const context = inject(containerContext);
```
amir20, line 3: If you search the code a little bit, you see `const { container } = useContainerContext();` is the right thing to use.

assets/composable/containerActions.ts
```diff
@@ -0,0 +1,47 @@
...
+      404: "container not found",
+      500: "unable to complete action",
+      400: "invalid action",
+    } as Record<number, string>;
+    const defaultError = "something went wrong";
+
+    actionStates[`${action}`] = true;
+
+    try {
+      await fetch(withBase(actionUrl), { method: "POST" }).then((response) => {
```
amir20, line 29: I think you missed my point earlier. Or maybe I am not getting it. But if you do `await` then you shouldn't need to do `.then(...)`.

Something like: 

```
const response = await fetch(...);
if(!response.ok){
...
}
```

Less code.

assets/composable/containerActions.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (!response.ok) {
+          showToast({ type: "error", message: errors[response.status] ?? defaultError, title: "Action failed" });
+        }
+      });
+    } catch (error) {
+      showToast({ type: "error", message: defaultError, title: "Container action failed!" });
+    }
+
+    actionStates[`${action}`] = false;
```
amir20, line 38: :)  `actionStates[action] = false;`

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,56 @@
...
+          <button
+            @click="stop"
+            :disabled="actionStates.stop || actionStates.restart"
+            v-if="container.state == 'running'"
+          >
+            <carbon:stop-filled-alt /> {{ $t("toolbar.stop") }}
+          </button>
+
+          <button
+            @click="start"
```
amir20, line 81: I am pretty sure this might be a typescript error and it should be `@click="start()"`.

internal/web/container_actions.go
```diff
@@ -0,0 +1,41 @@
...
+	container, err := client.FindContainer(id)
+	if err != nil {
+		log.Error(err)
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	err = client.ContainerActions(action, container.ID)
+	if err != nil {
+		log.Debugf("error while trying to perform action: %s", action)
```
amir20, line 33: Use `log.Errorf("error while trying to perform action: %v", err)` and remove the debug log. One log should be enough.

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,56 @@
           {{ $t("toolbar.show", { std: "STDERR" }) }}
         </a>
       </li>
+      <li class="line"></li>
```
amir20, line 68: I am confused here. Maybe I missed it last time. Doesn't this also need to be hidden?

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,56 @@
...
+      <!-- Container Actions (Enabled via config) -->
+      <ul v-show="enableActions">
```
amir20, line 70: Is this valid HTML? `<ul>` inside a `<ul>`?

amir20 (changes requested):
Sorry didn't have a lot of time to review this morning but did a quick round. Overall, looks a lot better. Hopefully we can merge today. I will be be back later in the day between meetings.

assets/composable/containerActions.ts
akash-ramaswamy: Yup! I refactored:)

assets/components/LogViewer/LogActionsToolbar.vue
akash-ramaswamy: I have changed it to `<template>` tag! so everything gets inside a single control flow!

assets/components/LogViewer/LogActionsToolbar.vue
```diff
@@ -65,17 +65,57 @@
           {{ $t("toolbar.show", { std: "STDERR" }) }}
         </a>
       </li>
+
+      <!-- Container Actions (Enabled via config) -->
+      <template v-if="enableActions">
```
amir20, line 70: Nice

nodis:
How do I enable the “start and stop containers” function?

akash-ramaswamy:
> How do I enable the “start and stop containers” function?

Hi @nodis Please refer this guide https://dozzle.dev/guide/actions
Feel free to ask for any doubts!

PR #2554, opened 2023-12-01
fix: changes --enableactions to --enable-actions

amir20:


akash-ramaswamy:
Hi @amir20 ! I just wanted to add up that I have missed to give a not in the docs saying this is from version 5.6.x

amir20:
Yea, I was thinking that. I think since it's disabled by default, its ok.

PR #2556, opened 2023-12-01
fix: Icon placement when hovering over containers with health checks.

ab623:
I noticed that when hovering over a container in the side list with a health check the spacing would be thrown off when the pin container button appears.

**Before**
Screenshot from 2023-12-01 17-18-48

I resolved this by fixing the spacing of the grid so that it stays aligned. Additionally I swapped the pin button to be after the heath icon. This is so that the button stays in a consistent position, allow the user to move down an continue to pin if they need to. 

**After**
Screenshot from 2023-12-01 17-33-56

### DIFF
--- a/assets/components/SideMenu.vue
+++ b/assets/components/SideMenu.vue
@@ -36,6 +36,7 @@
               <div class="truncate">
                 {{ item.name }}<span class="font-light opacity-70" v-if="item.isSwarm">{{ item.swarmId }}</span>
               </div>
+              <container-health :health="item.health"></container-health>
               <span
                 class="pin"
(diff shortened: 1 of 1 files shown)


assets/components/SideMenu.vue
```diff
@@ -145,7 +144,7 @@ const activeContainersById = computed(() =>
   @apply text-[0.95rem];
 }
 .containers a {
-  @apply auto-cols-[auto_max-content];
+  @apply auto-cols-[auto_max-content_max-content];
```
amir20, line 147: grid magic always blows my mind

amir20 (approved):

PR #2559, opened 2023-12-02
test: added test cases for container actions

akash-ramaswamy:
Hi @amir20 ! Sorry for delay, as i was looking at test packages documentation for reference 😅.

I have almost covered the critical functions.

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -149,7 +149,9 @@ func (d *Client) FindContainer(id string) (Container, error) {
 	return container, nil
 }
 
-func (d *Client) ContainerActions(action string, id string) error {
+func (d *Client) ContainerActions(action string, actionContainer Container) error {
+	id := actionContainer.ID
+
 	switch action {
 	case "start":
 		return d.cli.ContainerStart(context.Background(), id, types.ContainerStartOptions{})
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -149,16 +149,14 @@ func (d *Client) FindContainer(id string) (Container, error) {
 	return container, nil
 }
 
-func (d *Client) ContainerActions(action string, actionContainer Container) error {
-	id := actionContainer.ID
-
+func (d *Client) ContainerActions(action string, containerID string) error {
 	switch action {
 	case "start":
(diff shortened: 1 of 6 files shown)


internal/docker/client.go
```diff
@@ -149,7 +149,9 @@ func (d *Client) FindContainer(id string) (Container, error) {
 	return container, nil
 }
 
-func (d *Client) ContainerActions(action string, id string) error {
+func (d *Client) ContainerActions(action string, actionContainer Container) error {
```
amir20, line 152: Was there a reason you had to do this? I was looking at the other functions and all of them take `id`. So I think we should be consistent here and use `id`.  It also should make testing easier.

internal/web/routes_actions_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	handler := createHandler(mockedClient, nil, Config{Base: "/", EnableActions: true, Authorization: Authorization{Provider: NONE}})
+	req, err := http.NewRequest("POST", "/api/actions/start/localhost/123", nil)
+	require.NoError(t, err, "Request should not return an error.")
+
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+	assert.Equal(t, rr.Code, 200)
+
+	req, err = http.NewRequest("POST", "/api/actions/stop/localhost/123", nil)
```
amir20, line 36: Its generally bad practice to have multiple assertions in one test. It would be better to break up these tests where there is only one assert per test. This is also important because if one assert breaks, then the whole test will fail which is not great.

amir20 (commented):
I am away this weekend but did a quick review. Overall, good with some minor feedback.

internal/docker/client.go
akash-ramaswamy: Yeah, but I actually thought that I could use the result of `FindContainer`, since it would be an added check, that we are actually doing the actions on the intended container.

akash-ramaswamy: But anyways, I could use the `container.ID` property. So I would better take your suggestions :)

amir20: In this situation it might a premature optimization which breaks consistency and simplicity. So until we really need to pass the whole container as object, id should be enough.

PR #2562, opened 2023-12-04
fix: Updates to sidebar on mobile to align closer to desktop

ab623:
Updated the sidebar on the mobile to bring it closer to parity as the desktop. And fix some scrolling issues.

1. Updated the sidebar on mobile to show stopped containers
2. Sidebar either did not take the full height OR it overflowed and did not scroll on mobile. Both issues have been fixed by setting the height and overflow properties.

( FYI I hate CSS 😅 )

### DIFF
--- a/assets/components/common/MobileMenu.vue
+++ b/assets/components/common/MobileMenu.vue
@@ -20,7 +20,7 @@
     </div>
 
     <transition name="fade">
-      <div v-show="show">
+      <div v-show="show" id="side-container">
         <div class="mt-4 flex items-center justify-center gap-2">
           <dropdown-menu
             v-model="sessionHost"
@@ -39,7 +39,7 @@
 
         <ul class="menu">
           <li class="menu-title">{{ $t("label.containers") }}</li>
-          <li v-for="item in sortedContainers" :key="item.id">
+          <li v-for="item in sortedContainers" :key="item.id" :class="item.state">
             <router-link
               :to="{ name: 'container-id', params: { id: item.id } }"
               active-class="active-primary"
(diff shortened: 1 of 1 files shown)


ab623:
@amir20 Any advice on resolving the playwright failing test easily? Using playwright isn't the most strightforward process.

I assume the screenshots just need to be updated, but not sure how best to do that in the dev environments.

amir20:
I can probably just fix them on my side. Usually if you just delete them and run 'make int' will just update them.

assets/components/common/MobileMenu.vue
```diff
@@ -85,6 +85,15 @@ const sortedContainers = computed(() =>
 const hosts = computed(() => config.hosts.map(({ id, name }) => ({ value: id, label: name })));
 </script>
 <style scoped lang="postcss">
+#side-container {
```
amir20, line 88: I also hate css but ever since i used tailwind it has been better. 

Instead of using id with css, could you check if these are possible with tailwind. I am petty sure the overflow is possible.

ab623: The overflow as possible, and what I originally did. I just found it a bit weird to have the overflow as a tailwind class, and then attach the calc'ed high property as an id.

I just thought I would wrap them both up in the id.

ab623:
> I can probably just fix them on my side. Usually if you just delete them and run 'make int' will just update them.

You may need to do this on your side. 'make int' hangs for me with an error of not being able to connect to a remote TCP host. Haven't looked into it much, but as a docker compose it should "just work"

PR #2570, opened 2023-12-06
fix: fixes profile disk configuration when pinned is an empty set

amir20:


amir20:
@EDIflyer could you try this? It should be a quick fix to the empty pinned set. Try `amir20/dozzle:pr-2570`

EDIflyer:
No luck I'm afraid @amir20 still seems the same - have done a screen recording I'll send you in case helpful.

EDIflyer:
Hmm weird! I'm using Chrome and Firefox for the testing I'm doing. Glad the video at least showed the issue!

amir20:
Finally I was able to reproduce it! I had some previous data in localstorage so it was conflicting. I am a little unsure what is happening so I'll need to investigate.

amir20:
Alright, try `amir20/dozzle:pr-2570`. It turns out `[]` was being removed.  I think it should fix it now. But I am not sure if this is the best solution. So let's check first.

Issue #2571, opened 2023-12-06
Adding context menu for copying / actions on log messages


akash-ramaswamy:
**Is your feature request related to a problem? Please describe.**
Hi @amir20 ! I find it often very difficult to select the whole log row and copy. For example, if i need to copy this row, I would need to select the whole row.
![image](https://github.com/amir20/dozzle/assets/87078332/e9677489-0ede-4363-a920-779b31e55bab)

**Describe the solution you'd like**
Instead, If we could add a context menu on right-click of the current log row, we can provide options such as **copy log**, to make life easier.

Any thoughts on this?

amir20:
I had actually worked on this but scrapped it because the UI was ugly. If you can think of a minimal solution that doesn't interfere too much then I think it would be a good addition. 

For your example, you mean copy the JSON, right? So I guess it would just copy the original JSON without any filters?

akash-ramaswamy:
Thanks for reminding to improvise the *search menu* part too @amir20 ! This is an important point to make sure this is a really usable feature.

akash-ramaswamy:
@amir20 How about this hover action?


https://github.com/amir20/dozzle/assets/87078332/2af279ae-ee38-4a7d-bb99-d392e348143d

amir20:
I like it a lot. But I wonder what it would look like with `json`? Because there is another link that shows up to expand json. So it might not look good.  Have you thought of that?

amir20:
Also, not related, why is `DOZZLE_ENABLE_ACTIONS` showing a warning?

akash-ramaswamy:
> Also, not related, why is `DOZZLE_ENABLE_ACTIONS` showing a warning? 

I actually run two different containers, just was playing with old and new versions. Old version doesn't contain that feature right?

akash-ramaswamy:
> I like it a lot. But I wonder what it would look like with `json`? Because there is another link that shows up to expand json. So it might not look good.  Have you thought of that? 

Im planning to show this only on minimized view. Im thinking of something else for expanded view, or modify this accordingly.

Any thoughts?

amir20:
> I actually run two different containers, just was playing with old and new versions. Old version doesn't contain that feature right?

That's right. 

> This is just an initial idea. Im planning to show this only on minimized view. Im thinking of something else for expanded view, or modify this accordingly.

I don't like the current solution. So open to trying something else. I guess before spending time, it's worth calling out that you need a solution which: 

[...]

PR #2572, opened 2023-12-07
feat: actions on log messages

akash-ramaswamy:
Initial implementation of actions on log messages - Copy to Clipboard.


akash-ramaswamy:
#2571

amir20:
I reviewed the PR. A couple of comments: 

1. Seems like the hover is on top of the text not row. This is confusing because it seems to disappear easily. 
2. Some performance issues because you are using javascript to do hover. I'll comment on PR on this. 
3. Remembering why I didn't do this, I think you might need a better solution when the end of text would make the row jump down. I have attached a screenshot. 

Screenshot 2023-12-07 at 11 51 34 AM

With above example, when not hovered, it's one line. But when hovered, it is 2 lines.

assets/components/LogViewer/ComplexLogItem.vue
```diff
@@ -39,25 +51,18 @@ const [expanded, expandToggle] = useToggle();
...
+
+const isHovering = ref(false);
 </script>
 
 <style lang="postcss" scoped>
 .text-light {
   @apply text-base-content/70;
 }
-.fields {
-  &:hover {
```
amir20, line 49: This is how I was doing it before. When possible, _always_ prefer CSS over javascript.

assets/components/LogViewer/LogViewer.vue
```diff
@@ -15,7 +16,7 @@
       >
         <ic:sharp-find-in-page />
       </a>
-      <component :is="item.getComponent()" :log-entry="item" :visible-keys="visibleKeys" />
+      <component :is="item.getComponent()" :log-entry="item" :visible-keys="visibleKeys" :is-hovering="true" />
```
amir20, line 19: I don't understand what this is doing. It is hardcoded to `true`.

assets/models/LogEntry.ts
```diff
@@ -51,6 +51,10 @@ export class SimpleLogEntry extends LogEntry<string> {
   getComponent(): Component {
     return SimpleLogItem;
   }
+
+  copyLogMessageToClipBoard(): void {
```
amir20, line 55: This doesn't look like good logic to be part of LogEntry.ts. Models should hold data. Nothing more. Instead use the utils to do something like `function(data){...}`. Also you should look at https://vueuse.org/core/useClipboard/. It is a better solution and has a `copied` property that can be used to show a tooltip.

amir20 (changes requested):

akash-ramaswamy:
Hi @amir20 ! Thanks for the review. I will get back with the changes and improvements needed!

> 3. Remembering why I didn't do this, I think you might need a better solution when the end of text would make the row jump down. I have attached a screenshot. 

This is a valid point. I must also think of some way of solving this 🤔

assets/components/LogViewer/ComplexLogItem.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex gap-x-2">
+  <div class="flex gap-x-2" @mouseover="isHovering = true" @mouseleave="isHovering = false">
```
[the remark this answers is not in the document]
akash-ramaswamy: Thanks for this @amir20 , I didn't know that we can do this with tailwind. Im more fascinated by tailwind now.

assets/components/LogViewer/ComplexLogItem.vue
```diff
@@ -19,6 +19,18 @@
           <span class="font-bold" v-html="markSearch(value)" v-else></span>
         </li>
         <li class="text-light" v-if="Object.keys(validValues).length === 0">all values are hidden</li>
+        <li class="inline-flex w-fit align-middle" v-show="isHovering">
+          <div class="rounded px-1 py-1 hover:bg-slate-700 hover:text-secondary" @click="expandToggle()">
+            <carbon:row-expand v-if="!expanded" />
+            <carbon:row-collapse v-else />
+          </div>
+          <div
+            @click="logEntry.copyLogMessageToClipBoard()"
```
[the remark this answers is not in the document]
akash-ramaswamy: I tried a different solution now. I have moved the position of the copy button to end, making this stay constant.

assets/components/LogViewer/LogViewer.vue
akash-ramaswamy: It was a testing thing. Cleaned up!

assets/models/LogEntry.ts
akash-ramaswamy: yeah, I was doubtful when I was adding this. Thanks for the comment! I have isolated it to a separate component.

akash-ramaswamy:
Hi @amir20 I have moved the position of the `copy` button to the end, so that it  doesn't disturb the UI. And in future, we can make this a `menu` type vertical-three-dot menu, to make this a small dropdown to add more actions against log messages.


https://github.com/amir20/dozzle/assets/87078332/3ac9ae4c-bb84-4c80-a718-1f2f662e83b7

amir20:
Hi @akash-ramaswamy. This is going the right direction I think. I was actually thinking something similar where you move the icons to the right. One thought though, I noticed you have added padding to the right. This makes it so that there is less space. The most common use case is to look at logs, so we shouldn't sacrifice that width in my opinion. 

When I was thinking about this, I was thinking of using `position: absolute`. More examples https://tailwindcss.com/docs/position. Then do `right-0` and only show on hover. This should make it so that the icons show on _top_ of the content and therefore not take any space. 

Lastly, I feel like if there was a delay to show the icon would be great. But I believe that is not easily possible. You can probably use something like `transition` with delay. I noticed when moving the mouse the icon kept popping up and it felt weird to distracting. But this is minor.

akash-ramaswamy:
> Hi @akash-ramaswamy. This is going the right direction I think. I was actually thinking something similar where you move the icons to the right. One thought though, I noticed you have added padding to the right. This makes it so that there is less space. The most common use case is to look at logs, so we shouldn't sacrifice that width in my opinion.
> 
> When I was thinking about this, I was thinking of using `position: absolute`. More examples https://tailwindcss.com/docs/position. Then do `right-0` and only show on hover. This should make it so that the icons show on _top_ of the content and therefore not take any space.
> 
> Lastly, I feel like if there was a delay to show the icon would be great. But I believe that is not easily possible. You can probably use something like `transition` with delay. I noticed when moving the mouse the icon kept popping up and it felt weird to distracting. But this is minor.

I tried positioning the copy button. Its now on top of content as you are saying. But if content is more, its like overlapping a little bit, and I thing thats okay we can lower opacity of icon when not hovered on.. your thoughts @amir20 ?


https://github.com/amir20/dozzle/assets/87078332/d3ea499a-119d-4ebc-8f7d-d7c60337e66e

amir20:
Yea that looks better. You could also add a background color as the icon is pretty hard to notice on top of the text.

akash-ramaswamy:
> Yea that looks better. You could also add a background color as the icon is pretty hard to notice on top of the text.

I chosen the `text-primary`, since it looks cool over log row's background 😎

### Normal State
Screenshot 2023-12-09 at 4 32 56 AM

### Hovered State
Screenshot 2023-12-09 at 4 32 56 AM

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <div
+    class="absolute -right-1 flex min-w-[0.98rem] items-start justify-end align-bottom hover:cursor-pointer"
+    v-if="message.trim() != ''"
+    title="Copy Log"
+  >
+    <span
+      class="rounded bg-slate-800/60 px-1.5 py-1 text-primary opacity-0 transition-opacity delay-500 duration-1000 hover:bg-slate-700 group-hover/entry:opacity-100"
```
amir20, line 8: 1s duration is a bit much. Try 250. Most animation should under 250ms.

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -0,0 +1,35 @@
...
+
+const { showToast } = useToast();
+
+function copyLogMessageToClipBoard() {
+  navigator.clipboard.writeText(message);
+
+  showToast(
+    {
+      title: "Copied",
+      message: "Log message copied to clipboard",
```
amir20, line 29: don't forget to move these to locale

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -0,0 +1,35 @@
...
+
+<script lang="ts" setup>
+const { message } = defineProps<{
+  message: string;
+}>();
+
+const { showToast } = useToast();
+
+function copyLogMessageToClipBoard() {
+  navigator.clipboard.writeText(message);
```
amir20, line 24: Did you see https://vueuse.org/core/useClipboard/?

amir20:
Looking good <3 

how much work is left to finish it? There is also failed tests but i think you just need to update snapshot.

akash-ramaswamy:
> Looking good <3
> 
> how much work is left to finish it? There is also failed tests but i think you just need to update snapshot.

i updated the snapshot, but I find that `\\"` is added before every attributes, which was not there earlier.

amir20:
Make sure to rebase master. Vitest was updated.

amir20:
Make sure to rebase master. Vitest was updated.

assets/components/LogViewer/CopyLogMessage.vue
akash-ramaswamy: @amir20 I find that there is no option to use `i8n` inside `<script>`, am I not seeing properly?

amir20: Should be able to do `const { t } = useI18n();`

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -19,17 +19,21 @@ const { message } = defineProps<{
...
-
-  showToast(
-    {
-      title: "Copied",
-      message: "Log message copied to clipboard",
-      type: "info",
-    },
-    { expire: 2000 },
-  );
+  copy(message).then(() => {
```
akash-ramaswamy, line 26: Will update this file in a moment, with error handled

akash-ramaswamy: I think this is good to go!

akash-ramaswamy:
Hi @amir20 ! I have updated the PR with the cleanups.

akash-ramaswamy:
Yup, test cases are passing!

akash-ramaswamy:
Once again, thanks for your time on reviewing @amir20 :)

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div
+    class="absolute -right-1 flex min-w-[0.98rem] items-start justify-end align-bottom hover:cursor-pointer"
```
amir20, line 3: Right now this component can't be used in non absolute style because styles are hard coded. Instead of having the styles here, remove the ones that are not relevant to the component to where you use it. eg `<copy-log-message :message="..." class="absolute -right-1.../>`. You can probably keep `items-start justify-end...` since those are needed.

assets/components/LogViewer/CopyLogMessage.vue
amir20: Use `await` here. :) You should rarely need `.then()`.

assets/components/LogViewer/CopyLogMessage.vue
```diff
@@ -0,0 +1,39 @@
...
+    v-if="isSupported && message.trim() != ''"
+    title="Copy Log"
```
amir20, line 5: Should go in locale

amir20 (changes requested):
A couple quick fixes and ready to merge.

akash-ramaswamy:
I have at most extracted the classes which are not required for that `CopyLogMessage` component. Have a look @amir20 !

amir20:
I merged this. Do you have time fix the search mode? Or should I release first?

akash-ramaswamy:
> I merged this. Do you have time fix the search mode? Or should I release first?

Do you mean the log search mode @amir20 ?

amir20:
Yea, I actually found a bug. Looks like when the text is too long the icon is cut off. 

and back to the search, it probably should be all in one place before releasing. 

Screenshot 2023-12-09 at 6 21 53 AM

The bug might be minor because as you see in the screenshot, it's because I have a lot of text. But I have attached the logs in case you want to test it. 

[Imgproxy cache log Dec 9.gz](https://github.com/amir20/dozzle/files/13624386/Imgproxy.cache.log.Dec.9.gz)

amir20:
I was thinking you can put all the icons in one place and then using `v-if` dynamically adjust them.

amir20:
Oops more bugs when testing JSON...

Screenshot 2023-12-09 at 6 28 44 AM

akash-ramaswamy:
I will test this immediately and fix this asap @amir20 !

amir20:
I'll hold off releasing. JSON usage is high. 

For your testing, you can use https://github.com/amir20/echo which you can do `docker run -i amir20/echo < data.json` and pass any data you want. `amir20/echo` will randomly pick lines and print every second. 

You should try to test all the different variations. I am pretty on weekends, but try to respond when I can. 

I have created a bug so we don't keep discussing here.

Issue #2579, opened 2023-12-09
Copy with JSON logs is overlapping 


amir20:
Oops more bugs when testing JSON...

<img width="1389" alt="Screenshot 2023-12-09 at 6 28 44 AM" src="https://github.com/amir20/dozzle/assets/260667/2237f1bf-8555-48d6-ba0a-7d691427b33d">

_Originally posted by @amir20 in https://github.com/amir20/dozzle/issues/2572#issuecomment-1848425449_

amir20:
@akash-ramaswamy created this bug to continue discussion. Let's bring all related bugs you find here.

akash-ramaswamy:
Hi @amir20 , regarding this JSON bug keeping the parent `div` (which wraps timestamp, stdout/stderr, JSON and copy button) if set to `w-full` it resolves this bug.

akash-ramaswamy:
Hi @amir20 , regarding this comment https://github.com/amir20/dozzle/pull/2572#issuecomment-1848424922 this is fine. I was initially thinking this too!

I thought we also could add `word-break` to the log message div, as it could break into new line automatically and width is maintained!

amir20:
Send a pr and I'll test it. I have a complicated set up so can test edge cases.

akash-ramaswamy:
@amir20 raised a PR

amir20:
Merged the other PR. So looks like search is just remaining. 
<img width="1406" alt="Screenshot 2023-12-09 at 6 21 53 AM" src="https://github.com/amir20/dozzle/assets/260667/3379609f-f3f2-4d21-96c8-48d27ee27b3f">

amir20:
Tested #2580. Looks good!

akash-ramaswamy:
> Merged the other PR. So looks like search is just remaining.

I have a query though, after we did this change how would a user choose `Jump to Context` if the search entry is at the top?

![image](https://github.com/amir20/dozzle/assets/87078332/e67075b7-5a6d-4e7a-b83f-2a4c91f65a45)

amir20:
I think it would be fine since the user can scroll. But I guess it will be one of those things I'll have to test out to really know.

akash-ramaswamy:
> I think it would be fine since the user can scroll. But I guess it will be one of those things I'll have to test out to really know.

But the image I highlighted is the **First Log entry** 🥶

amir20:
Maybe for now it will be ok since selecting only the first log won't be common. 

[...]

PR #2580, opened 2023-12-09
fix: actions on log UI issues

akash-ramaswamy:
#2579

### DIFF
--- a/assets/components/LogViewer/ComplexLogItem.vue
+++ b/assets/components/LogViewer/ComplexLogItem.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative flex gap-x-2">
+  <div class="relative flex w-full gap-x-2">
     <div v-if="showStd">
       <log-std :std="logEntry.std"></log-std>
     </div>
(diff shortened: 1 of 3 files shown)


amir20:
Won't have time to test this until later. But perhaps you can also include the fixes to search more in one PR?

akash-ramaswamy:
> Won't have time to test this until later. But perhaps you can also include the fixes to search more in one PR?

this solves that issue too, which you mentioned in this https://github.com/amir20/dozzle/pull/2572#issuecomment-1848424830

amir20:
How does this fix the search? That's an icon on the left that I am saying should be moved to the right with the other icons. Can you send a screenshot?

akash-ramaswamy:
> How does this fix the search? That's an icon on the left that I am saying should be moved to the right with the other icons. Can you send a screenshot?

Oh sorry that you misunderstood! I was saying that the copy button will not be facing issues anymore, since i added `word-break` property.

akash-ramaswamy:
This is after the fix applied for JSON bug

image

amir20:
Right. As part of this fix can you also moved the jump to context button to the right so it's consistent?

amir20:
It does look like the JSON bug is fixed now. 🎉

akash-ramaswamy:
> Right. As part of this fix can you also moved the jump to context button to the right so it's consistent? That was one of the original requirements. 😁

I think moving this would require little more work in UI too, As i want to make the button same as copy-button style :) ? could the release be kept hold until then?

akash-ramaswamy:
> Right. As part of this fix can you also moved the jump to context button to the right so it's consistent? That was one of the original requirements. 😁

Also I thought that this could be a separate issue, or linked to #2571 ?

amir20:
I can merge this soon but I think I'll hold off to release. Search feature is used a lot so it's probably best to be consistent before releasing.

PR #2583, opened 2023-12-10
fix: disables hover on mobile

amir20:


amir20:
@akash-ramaswamy with this change, `:hover` is disabled on mobile. So just heads up.

PR #2584, opened 2023-12-10
feat: Moves Jump-To-Context right

akash-ramaswamy:
Hi @amir20, PR raised for #2579


amir20:
Wow bigger change than I thought. I'll review later.

amir20:
This is great. Works perfectly and the code is done just right. I was debating if `:message` should be used from `logEntry` but it's different for JSON vs simple. 

Going to merge. Test a little and release. 

🎉

akash-ramaswamy:
> This is great. Works perfectly and the code is done just right. I was debating if `:message` should be used from `logEntry` but it's different for JSON vs simple. 

Hi @amir20 ! The UI side was little simple because of your previous input *to extract un-relevant classes from the component itself* 😄

Issue #2601, opened 2023-12-17
Avatar not showing up correctly 


amir20:
> > Have updated to the latest release and it's not a broken image anymore but I get a '12' for some reason?
> 
> That is weird for sure. I can take a guess. Can you let me know what `Name:` you are using? Paste your `users.yml` without password so I can test it? You can email it to me too.

PS further to my email found from another bug that it was `/data` that I had to map to access config files. Have done so and after it initially being empty I changed a setting about found a `profile.json` in `data/EDIflyer` however no `users.yml`.  I'm not using user management though, just `--forward-proxy` - not sure if that's related?  It works find for the sites where the Authelia user has an email addresses known to Gravatar but not for this one (which isn't a known email).  No longer a broken image but this weird '12'!

_Originally posted by @EDIflyer in https://github.com/amir20/dozzle/issues/2591#issuecomment-1859149044_

amir20:
This should be fixed in master. Will release a build today.

PR #2602, opened 2023-12-17
fix: escapes avatar url correctly. see #2601

amir20:


amir20:
@EDIflyer can you try this? `amir20/dozzle:pr-2602`.

While I wasn't able to reproduce your problem exactly, I did find another bug. 

If the issue still happens, then please enable debug logs and send me the log that says `Fetching avatar from...".

Thanks!

EDIflyer:
Yep still happening - 
`level=debug msg=&quot;Fetching avatar from https://gravatar.com/avatar/617016fc8b09804be4fbfa5eb2054ac9?d=https%3A%2F%2Fui-avatars.com%2Fapi%2F//128&quot;`

When I try that I get the 12 image that you're then showing.

When I look up the email address that's being passed through from Authelia (that you correctly pick up on hover) - I get this error on the Gravar.com website, which is what I would expect as I haven't registered that address with them...
image

amir20:
I see what the problem is now. According to that URL, your displayname is empty? Can you confirm what Authelia is passing to Dozzle for `Remote-Name`? 

I don't have the exact set up as you but I'll try to look at documentation. 

> When I look up the email address that's being passed through from Authelia (that you correctly pick up on hover) - I get this error on the Gravar.com website, which is what I would expect as I haven't registered that address with them...

Yes, it should just show your initials for your name when Gravatar doesn't exist.  For you it would be AR. Which ironically would same initials as me.

EDIflyer:
Authelia is setup with `displayname` - I noticed in the dB for this is wasn't in quotes but was for the other site - I tried changing it but no change.  Not sure where I can see what `Remote-Name` is passed to Dozzle as it doesn't appear in the debug logs?

amir20:
I don't print it. But you can use that other app you had found to print all the headers. Let me know if it's easier for me to add logs.

amir20:
@EDIflyer I think I found the problem. I found your old comment https://github.com/amir20/dozzle/issues/2437#issuecomment-1774121363 which shows your setup doesn't have `Remote Name`. If you remember we were actually discussing this.

amir20:
OK test with `amir20/dozzle:pr-2602` again. Now if name is blank I use username instead.

EDIflyer:
Yep that makes more sense now - it's got the first couple of characters of the username...
image

amir20:
The 12 is from 128 size I pass to the avatar service. Since the name was empty it thought username is 128. 

Finally figured this out. 🤓

EDIflyer:
Haha love it @amir20 - shame it wasn't `[42](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy)`😂

Issue #2619, opened 2023-12-22
Improve Log Line Copy Button


nrgbistro:
When copying a log line using the copy button on the right, the result in my clipboard contains additional character codes that make searching for my issue harder than expected. For example, I have an error from one of my containers. Here is what's copied using the button:

`[31m[Nest] 7  - [39m12/22/2023, 9:56:31 PM [31m  ERROR[39m [38;5;3m[MediaService] [39m[31mAn error occurred while configuring transcoding options: Error: ENOENT: no such file or directory, scandir &apos;/dev/dri&apos;[39m`

Here is what I would have expected the copy button to produce:

`[Nest] 7  - 12/22/2023, 9:56:10 PM   ERROR [MediaService] An error occurred while configuring transcoding options: Error: ENOENT: no such file or directory, scandir '/dev/dri'`

amir20:
It's because you have interactive mode with colors. Maybe something Dozzle can detect and remove? @akash-ramaswamy any ideas?

akash-ramaswamy:
Hi @amir20 ! I have tested this fix, hope this resolves the issue :)

amir20:
Merged the fix in master. Thanks @akash-ramaswamy !

amir20:
v5.8 will address this.

PR #2621, opened 2023-12-23
fix: copy log ansi character parsing and html decode

akash-ramaswamy:
fix for #2619

### DIFF
--- a/assets/components/LogViewer/SimpleLogItem.vue
+++ b/assets/components/LogViewer/SimpleLogItem.vue
@@ -9,7 +9,7 @@
     ></div>
     <log-message-actions
       class="duration-250 absolute -right-1 opacity-0 transition-opacity delay-150 group-hover/entry:opacity-100"
-      :message="logEntry.message"
+      :message="decodeXML(stripAnsi(logEntry.message))"
       :log-entry="logEntry"
     />
   </div>
@@ -17,6 +17,8 @@
 <script lang="ts" setup>
 import { SimpleLogEntry } from "@/models/LogEntry";
 import AnsiConvertor from "ansi-to-html";
+import stripAnsi from "strip-ansi";
+import { decodeXML } from "entities";
 
 const ansiConvertor = new AnsiConvertor({ escapeXML: false, fg: "var(--text-color)" });
 defineProps<{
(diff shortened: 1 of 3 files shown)


assets/components/LogViewer/SimpleLogItem.vue
```diff
@@ -9,14 +9,16 @@
     ></div>
     <log-message-actions
       class="duration-250 absolute -right-1 opacity-0 transition-opacity delay-150 group-hover/entry:opacity-100"
-      :message="logEntry.message"
+      :message="decodeXML(stripAnsi(logEntry.message))"
```
akash-ramaswamy, line 12: Hi @amir20 ! I did this here and not in the `LogMessageActions` component itself, since the component handles `JSON` string also.

akash-ramaswamy (commented):
Hi @amir20 ! I have tested this fix against some ansi color coded log strings. It works and I didn't find any issue. Also i chained `decodeXML` to make sure the string is copied as it is and not with any `html` entities.

I tried of just simply copying the `textContent` of the div, but that would require us to have some kind of identifier for each log entry div, which I felt would be unnecessary at this time.

amir20:
Nice! I am traveling this week but maybe merge tomorrow via phone and testing can happen on master.

amir20:
Looks good. Merging. 

Strip colors seems like a good idea. 

So many people complain about colors not looking right that maybe Dozzle should just disable colors and enable as an option. Something to do in the future.

assets/components/LogViewer/SimpleLogItem.vue
amir20: I was just thinking of something. This might be a performance issue since it is being computed even user doesn't need to copy. 

Although, probably tiny and just not a big deal. Something to measure before releasing.

akash-ramaswamy: > I was just thinking of something. This might be a performance issue since it is being computed even user doesn't need to copy.

@amir20 , maybe we can implement this in the `LogMessageActions` widget itself?

amir20: I think it's a micro optimization. But if we wanted to it can be passed as a lazy function. Then with toValue converted to real value. 

We should measure first before doing too much work.

PR #2626, opened 2023-12-27
feat!: strips all ansi colors and disables showing any colors

amir20:


amir20:
@akash-ramaswamy I like what you did in the other PR. I think the right solution was to always strip all ansi colors. So I have done that here. Now all colors are removed in API. 

Can you take a look? Thanks!

akash-ramaswamy:
Yeah @amir20 ! This is the right decision. But do you have any plans to provide this optionally? or completely remove colors?

akash-ramaswamy (commented):
Doing this from backend itself will lower the perf. overhead in frontend. This is great choice @amir20 !

amir20:
Thanks! 

> But do you have any plans to provide this optionally? or completely remove colors

I thought about this a lot. Looking back, there were a few issues related to colors. Most people wanted to remove the colors because their colors conflicted with dark mode (or whatever was shown on the HTML). When thinking about it, I think it makes sense to respect the colors in the browser. 

What do you think? I am not sure how many people would actually want the colors. It seem to cause more headaches.

akash-ramaswamy:
> What do you think? I am not sure how many people would actually want the colors. It seem to cause more headaches. 

Hmm, yes. You are right! Now I feel the same way too.

PR #2796, opened 2024-02-27
feat: adds hideMenu as a query parameter

amir20:


clabnet:
It work, go ahead. 
Thank you @amir20

PR #2809, opened 2024-03-05
docs: add missing base slash and remove non-existent nginx configuration

pothitos:
Thank you, Amir, for your awesome work!

I submit the two necessary modifications I made in order to provide Dozzle in a subroute of my site.


amir20:
Nice. Merging!

PR #2893, opened 2024-04-13
feat: groups containers by stack or compose when possible

amir20:
- [x] Set the order of grouped containers to alphabetical 
- [x] Keep track of open/closed details 
- [x] Hide empty groups 
- [x] Show stopped containers as different color 
- [x] Make tests pass


dhellwich:
@amir20 Hi, currently if a stack has only one container, the stack name isn't shown in Dozzle. The one container in the stack is placed under "All Containers" category. Is it possible to show the stack name if only one container is running inside the stack?

amir20:
@dhellwich 

This was discussed a lot before I had to choose one. Many people didn't want a single group because vertically it was getting long. 

I got a lot going on with `dtop` these days so won't have time to work on this feature. But you can send a PR if you wish. It makes sense to be a property in the settings.

PR #2948, opened 2024-05-12
feat: Adding debug log when host-specific certificate path not found

fredskis:
Ran into this while figuring out why my remote host config wasn't working. Debug logging didn't tell me where it was expecting (and failing) to find the client certificate files, only that it didn't find them at the default `/certs` location. 

Existing debug log shows the final default path it looks for, but not the host-specific one it tries first. 

Added a debug log action to specify when this fails, where it expects the files to be.


amir20:
Thanks!

PR #2968, opened 2024-05-22
fix: Coerce log level to lowercase when guessing in structured json

rm-hull:
My project generates structured JSON log messages with the `level` attribute having values of DEBUG, INFO, WARN, ERROR, etc <-- all in upper case. When viewed in dozzle, the purple/green/orange/red marker dots don't show for these entries. 

I tracked this down to the **level_guesser.go** file and noticed some missed test coverage. This PR should address that deficiency.


amir20:
Nice find!

Issue #3052, opened 2024-06-21
💁 Feedback request on agent support for swarm and remote hosts 


amir20:
## Why is agent needed? 

Dozzle doesn't support swarm mode natively. Remote hosts are done by setting up connections with TLS or https://github.com/Tecnativa/docker-socket-proxy. However, there has been many problems with this approach. Here is some I have found: 

1. Security - `docker-socket-proxy` is insecure.  Setting up TLS is the preferred route, but doing so is very complicated. It requires generating certs, updating docker settings and listening to TCP port. 
2. Ease of use - The majority of bugs I get are related to not being able to setup `socket-proxy`. Dozzle works as design but people find the two might break. 
3. Stability - Dozzle uses the native Docker API which handles connectivity. When a connection fails, Dozzle doesn't try to reconnect. A lot of folks have wanted to reconnect. 

Setting up an agent can solve many problems. 

## How will the agent be implemented? 

Currently, I am working on an agent using gRPC. It will use TLS by default. Compose file for swarm might look like: 

```yml

services:
  dozzle:
    image: amir20/dozzle
    ports:
      - "8080:8080"
    environment:
      DOZZLE_MODE: swarm
    deploy:
      mode: global
     endpoint_mode: dnsrr
```

In this `swarm` mode, I want minimum setup. Dozzle will be smart enough to create it's own mesh using `endpoint_mode: dnsrr`. 

For remote hosts, it could look like this: 

```yml
services:
  dozzle:
    image: amir20/dozzle
    ports:
    environment:
      DOZZLE_REMOTE_AGENTS: host1,host2   
```   
 
Where the agents are setup remotely like this: 

```yml
services:
  agent:
    image: amir20/dozzle
    ports:
      - 7007:7007
    environment:
      DOZZLE_MODE: agent
```

In this mode, the agent will expose its own API over port 7007. Then there will be one installation of Dozzle UI which will connect to all agents. 

## Deprecation of remote connections 
I think with the implementation of agents, it would be safe to remove remote connections. However, I am not sure if people still care to use it. I can't imagine a use where remote connection with `socket-proxy` is preferred over Dozzle's own agent. 

## Questions on top of my head
1. How do the sample compose files look? Are they easy enough to use? 
2. How much do folks care about security? 
3. Should the use of `socket-proxy` be removed in the future? 
4. What are some example of wanting to keep remote connections with agent support? 

Feedback welcomed!

githubbiswb:
[...]

PR #3058, opened 2024-06-26
feat!: implements swarm mode with agents

amir20:
This pull request introduces a significant change in design by implementing Dozzle agents. Dozzle agents communicate with each other over a secured RPC. Only Dozzle agents can talk to each other. This is achieved by using pinned certificates. 

There are two use cases for agents. The first one is to support Dozzle in Swarm mode without much configuraiton. The second use case is to replace remote hosts. 

## Dozzle in Swarm Mode 

In Swarm mode, Dozzle creates connection to a cluster of agents running together. In this mode, all logs are streamed to the UI. There isn't much configuration needed except `DOZZLE_MODE=swarm`. 

```
services:
  dozzle:
    image: amir20/dozzle:pr-3058
    environment:
      - DOZZLE_MODE=swarm
    volumes:
    ports:
      - "8080:8080"
    networks:
      - dozzle
    deploy:
      mode: global

networks:
  dozzle:
    driver: overlay
```

Note `amir20/dozzle:pr-3058` is required to point to this PR. 

You can see agents working together by doing `docker service logs -f  <service_name>`. 

## Setting up remote host 

For remote host, it works similar to swarm mode except that the agent's endpoints need to be configured. 

To start an agent, you need to run `docker run --volume=/var/run/docker.sock:/var/run/docker.sock -p 7007:7007 amir20/dozzle:pr-3058 agent`. This will start an agent and starting listening to port `7007`. 

Then you can start Dozzle as usual `docker run  -p 8081:8080 amir20/dozzle:pr-3058 --remote-agent [ip]:7007`. 

This PR is still in development. Any testing would be appreciated. 

See #3052 for more details. 

Still to do: 

- [x] Fix healthcheck 
- [x] Fix beacon on page load 
- [x] Add a simple int test 
- [ ] Test actions


githubbiswb:
Looks good to me, and works very well. A much simpler config and less images, always a plus

Keep up the great work!

devioa:
docker build error
```
 => ERROR [builder 10/14] COPY shared_key.pem shared_cert.pem ./                                                 0.0s
------
 > [builder 10/14] COPY shared_key.pem shared_cert.pem ./:
------
Dockerfile:49
--------------------
  47 |     COPY main.go ./
  48 |     COPY protos ./protos
  49 | >>> COPY shared_key.pem shared_cert.pem ./
  50 |
  51 |     # Copy assets built with node
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref e1af3785-ab55-4be1-9acc-2344d09c5acf::k0alhi681gfc2txlmd9df3jeq: "/shared_key.pem": not found
```
Do I need to generate this certificate myself?

amir20:
Yep. make generate will create them. Official releases have private keys that are secret. 

Use make docker to build.

devioa:
> Yep. make generate will create them. Official releases have private keys that are secret.
> 
> Use make docker to build.

Thanks

Enissay:
Thank you, that's a very nice feature! 

I have never managed to make the previous way using ssh to work, so I ended up deploying many instance each with own port/address, but it was still a pain if a container is deployed randomly...

> Due to implementation details, the name for the service must be exactly dozzle. This is because Dozzle uses the service name to discover other Dozzle instances in the swarm. In the future, it would make sense to query the Docker API for all services with the current container ID.

Would it be possible to pass the service name as a variable ? that will make it more flexible...

amir20:
@Enissay Can you send a pr if you know how? I won't be able to get this for a while.

Enissay:
> @Enissay Can you send a pr if you know how? I won't be able to get this for a while.

Thank you, but I am afraid Go and Vue are out of my league, so I guess that can wait xD

amir20:
@Enissay I think I want to find a way to discover the service name. I know there is a way so going to find that way for now.

Enissay:
> @Enissay I think I want to find a way to discover the service name. I know there is a way so going to find that way for now.

You mean through docker api itself ? If you can tell me what you have in mind, I can explore it for you...
One dumb way I use for some similar aliases would be `docker service ls | grep amir20/dozzle | awk '{print $2}'`, but that would fail if more than one services are available :x

amir20:
I think it would be something like 

```
 docker inspect <id> | jq '.[0].Config.Labels["com.docker.swarm.service.name"]'
```

`<id>` could be coming from `echo $HOSTNAME`.  But I am really not sure if that is true in all use cases.

Enissay:
I tried in a couple containers using `docker inspect "$(echo $HOSTNAME)" | jq '.[0].Config.Labels["com.docker.swarm.service.name"]'` and it indeed returned the correct service name.

PR #3090, opened 2024-07-10
Update README.md

bluepuma77:
Removed all `:ro` from `/var/run/docker.sock:/var/run/docker.sock`, as they do not work on sockets and imply non-existent security


amir20:
I don't understand why the int tests are failing. I am going to merge anyway and hope for the best.

Issue #3189, opened 2024-08-09
Ability to customize a container's name through a Docker label


Jeto143:
### Describe the feature you would like to see

In some cases, setting the name of the container through Docker (via `--name` or `container_name`) is not possible. For instance, using [Coolify](https://coolify.io/), you can customize container names but if you do so, you lose the rolling update feature, so container names usually are randomized slugs.

### Describe how you would like to see this feature implemented

Since we already have [`dev.dozzle.group`](https://dozzle.dev/guide/container-groups#custom-groups) as a possible label to define the name of a group, why not introduce a `dev.dozzle.name` label as well, that if defined for a given container, would supersede its actual name within Dozzle?

amir20:
Not a bad idea. Seems like a simple a change if you want to send a PR.  Changes would be at: 

https://github.com/amir20/dozzle/blob/ffc4559047f52a8e0ac3f31cf30ecdc24be1c293/internal/docker/client.go#L358

https://github.com/amir20/dozzle/blob/ffc4559047f52a8e0ac3f31cf30ecdc24be1c293/internal/docker/client.go#L382

Jeto143:
Thanns! I'll give it a shot ASAP.

amir20:
Any update? I won't have much time either so I might close if neither of us can do it. 🙈

Jeto143:
> Any update? I won't have much time either so I might close if neither of us can do it. 🙈

Still planning on it, I'll give it a go either this week or next week.

Jeto143:
@amir20 There you go, lmk what you think!

amir20:
Nice! Merged. I am hoping you tested it :)  It is now in master so if you give it a try with `amir20/dozzle:master` and test it. I can release later today.

Jeto143:
> I am hoping you tested it :)

I did a basic test, yeah: I had a bunch of containers already running, then ran a quick `docker run` command after the change with the new label and saw the new container appear on the list with the desired name.

PR #3197, opened 2024-08-13
Implement support for custom container names

Jeto143:
This introduces the ability to customize a container's name by adding a `dev.dozzle.name` label to it.

Related issue:
https://github.com/amir20/dozzle/issues/3189

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -65,6 +65,7 @@ export default defineConfig({
           { text: "Actions", link: "/guide/actions" },
           { text: "Agent Mode", link: "/guide/agent" },
           { text: "Changing Base", link: "/guide/changing-base" },
+          { text: "Container Names", link: "/guide/container-names" },
           { text: "Container Groups", link: "/guide/container-groups" },
           { text: "Data Analytics", link: "/guide/analytics" },
           { text: "Display Name", link: "/guide/hostname" },
--- a/docs/guide/container-names.md
+++ b/docs/guide/container-names.md
@@ -1,0 +1,30 @@
+---
+title: Container Names
+---
+
+# Container Names
+
+By default, Dozzle retrieves container names directly from Docker. This is usually sufficient, as these names can be customized using the `--name` flag in `docker run` commands or through the `container_name` field in Docker Compose services.
+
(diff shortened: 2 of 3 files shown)


amir20:
Nice. And updated docs too 🚀

amir20:
@Jeto143 quick favor. When developing Dozzle for the first time, were the instructions at https://github.com/amir20/dozzle?tab=readme-ov-file#building enough? Let me know if that should be updated.

Jeto143:
> @Jeto143 quick favor. When developing Dozzle for the first time, were the instructions at https://github.com/amir20/dozzle?tab=readme-ov-file#building enough? Let me know if that should be updated.

Sure! First a quick disclaimer, I have never developed in Go before (even though I've wanted to learn it for quite some time). I use MacOS and IntelliJ IDEA.

The process overall went well but I had to do a few things that weren't mentioned in the README:

1. Had to add `$(go env GOPATH)/bin:$PATH` to my PATH after installing reflex. This shouldn't need to be added to the README as I'm sure it's common knowledge and not specific at all to this project.
2. Had to [enable Go modules](https://www.jetbrains.com/help/idea/2024.2/create-a-project-with-go-modules-integration.html#enable-go-modules-in-a-project) after importing the project inside IntelliJ so that it could understand the `github.com/*` imports. This is IDE-specific and shouldn't need to be added either.
3. I had to manually run `make build` (`vite build`) once, otherwise the startup process (`make dev`) would fail with `[backend] [00] main.go:28:12: pattern all:dist: no matching files found`. This one might need your attention as I'm assuming we shouldn't need to build assets manually during development. If you have a `dist` folder locally and remove it you might be able to reproduce this.
4. When running `make dev`, my browser opens automatically on http://localhost:5173 but the page sometimes fails to load ("localhost refused to connect") for a few seconds (I guess until it can load all the containers etc). It's really no big deal at all, but the first time it happened I didn't notice and thought something was wrong.

Hope this'll help!

amir20:
Ah nice. That's good. I'll update `make dev` to also depend on auto generated files. 

On 4, that's because the browser is opening but the Go binary is still compiling. I am not sure of a better solution.

PR #3205, opened 2024-08-14
bug: Translated incorrect into German correctly and added missing

Mephisto090:


Mephisto090:
I have added some that were in English but not in German

amir20:
Nice!

amir20:
@Mephisto090 Try `amir20/dozzle:master` You should see the new changes.

PR #3206, opened 2024-08-15
feat: Translate the example log entries

Mephisto090:
Please check again whether I have included the variables correctly. I have never worked with `.vue` files before. The log has already been translated into German. I have not touched the other languages.

### DIFF
--- a/assets/pages/settings.vue
+++ b/assets/pages/settings.vue
@@ -171,14 +171,14 @@ const hoursAgo = (hours: number) => {
 };
 
 const fakeMessages = [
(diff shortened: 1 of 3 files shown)


amir20:
Looks, right. Were you able to actually test it? You can do it with `make dev`.

amir20:
Tested it. Looks good to me.

Mephisto090:
You'll have to explain the teten thing to me in more detail, then I'd like to try it out next time. :)

amir20:
Since this is a side project for me I don't have a lot of time on it. You have to bring up Dozzle and instructions are at https://github.com/amir20/dozzle?tab=readme-ov-file#building. 

It does require being comfortable with building and compiling code. So if you are going to send a lot of PRs then you should figure out how to test the code locally.

Issue #2432, opened 2023-10-20
Add filter for log levels


Corbie-42:
The log level detection is already working fine. Now it would be great, if we could filter the output by log level. Every level should be filterable individually.

amir20:
Yea, that's a good idea. But there is an issue why I didn't implement this. Right now, I _guess_ the level.  https://github.com/amir20/dozzle/blob/021d3f7a9938d81878d9ad71ec2dc36384f7e2a6/docker/level_guesser.go#L11

I use a combination of regex and filters to check for level. It could totally be wrong.  I haven't implemented this because I don't think I am always right. So what should happen to the filtering when it can't find the level? 

Example would be say there are 4 lines, WARN, WARN, INFO, NO_LEVEL. If I filter by WARN, should it not show NO_LEVEL too?

EDIflyer:
Perhaps rather than treating as levels (as proper logging would, where you'd see that level and anything higher but nothing lower) you could instead treat them more like custom filters in Excel where the user could pick any combination of them that they want by ticking/unticking them? Hope that makes some sense!

Corbie-42:
Maybe you can treat NO_LEVEL just like the level detected above. This often happens, if a log entry consists of multiple lines, like a stack trace.

You could also make it a option to treat such entries differently and show/hide them on demand.

@EDIflyer that's what I meant 😊

amir20:
I think it would be weird if someone is filtering by INFO and then WARN doesn't show up. What do other tools do? Does anybody have examples?

Corbie-42:
You can have a look at the [Chrome developer console](https://developer.chrome.com/docs/devtools/console/log/#level). There the log levels can be selected individually.

EDIflyer:
Ah yes, good idea of an example!
![image](https://github.com/amir20/dozzle/assets/13610277/760a03b4-0ab5-4c8f-ae4d-a2812aa75dcd)

amir20:
This is the javascript console right? I think with JS, you can only have `console.log`, `console.error` and `console.warnings` so that makes sense. 

But perhaps this is good enough and one option could be undefined levels. 

I'll keep thinking about this. Gotta finish the mult-user authentication first. :)

EDIflyer:
Yep just the standard console in Chrome - It covers non JS things too (CSS, etc.) that might be why it has those other levels...
[...]

PR #3237, opened 2024-08-27
feat: adds a side panel for complex logs that enables reordering and enabling fields 

amir20:
Screenshot 2024-08-27 at 11 26 30 AM


amir20:
@FrancYescO I saw you reacted on this PR. I am merging it soon. I think it is the last thing I want to do before releasing. Could you test `master`?

FrancYescO:
just got a glitch: if a value is pretty long it will hide the togglebox (will go on the left out of screen and you need to horizontal scroll to show)

FrancYescO:
also consider adding a label "drag to reorder" or some icon that will show the cross cursor, otherwise will be a very hidden feature

FrancYescO:
maybe in the time on top of the sidebar you can add also the ms of the log datetime

amir20:
Thanks @FrancYescO 

Yea there are some horizontal spaces that need to be improved. And even more on mobile. 

> also consider adding a label "drag to reorder" or some icon that will show the cross cursor, otherwise will be a very hidden feature

Doesn't the cursor already change to a cross? Or do you mean additionally put a note about it? 

> and idk if is and easy add also a toggle for the specific log level (but than i have some doubt on how this can be toggled back to be visualized)

What does this mean? Do you mean filter by log levels? https://github.com/amir20/dozzle/issues/2432

FrancYescO:
Yes but it changes only while dragging, there is no indication that you are allowed to drag 😃 

And yes i was referencing to the same feature of the #2432

redge76:
It's an interesting feature but when there is a LOT OF fields (like the access log of traefik) then if is impossible to see all fields on one page (and I'm on a 32" monitor...)
Previously you cloud click on a line and get all the fields concatenated. It was hard to read but at least you could see everything without scrolling too much.
I don't know what would make a good solution to this issue. Maybe a tooltip with a condensed line as previously.

amir20:
Screenshot?

redge76:
There is not much to see. The traefik access log had 40 fields. if you enable a few and click on a line you get this:
image

Previously when you clicked on a line you had the same result you have now when you select all fields (where you could select the fields you wanted to display)

amir20:
I don't understand what you are saying is not possible now. If it's a bug then open an issue with better description. I have read your comment a few times and can't make any sense of it.

Issue #3240, opened 2024-08-29
Searching logs only works sometimes


WeetbixNMilk:
### 🔍 Check for existing issues

- [X] Completed

### How is Dozzle deployed?

Standalone Deployment

### 📦 Dozzle version

v8.2.1

### ✅ Command used to run Dozzle

A Vanilla install of the standard Dozzle container for Unraid. https://hub.docker.com/r/amir20/dozzle/

version: "3"
services:
  dozzle:
    container_name: dozzle
    image: amir20/dozzle:latest
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
    ports:
      - 8888:8080

### 🐛 Describe the bug / provide steps to reproduce it

Searching for a term will sometimes load infinitely, return a result, not return any results or hang. This is all on the same instance, without any restarts, changes or modifications. The search window is opened with Ctrl + F, issues occurs when both pasting and typing text. 

I will attach screenshots, each image is taken consecutively, only clicking into a different container, and then back to the original container logs in between screengrabs. (I am aware the search just looks like it is loading, but it does this infinitely. Sometimes it loads immediately without even restarting the logged container).

![Screenshot 2024-08-29 115021](https://github.com/user-attachments/assets/b6c98cff-6dde-4303-9b82-cbe0980edc24)
![Screenshot 2024-08-29 115202](https://github.com/user-attachments/assets/388bb799-b899-471b-83f7-ae0bae459e45)

### 💻 Environment

Client:
 Version:    24.0.9
 Context:    default
 Debug Mode: false

Server:
 Containers: 56
  Running: 43
  Paused: 0
  Stopped: 13
 Images: 50
 Server Version: 24.0.9
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 nvidia runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7c3aca7a610df76212171d200ca3811ff6096eb8
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
[...]

PR #3245, opened 2024-09-01
feat!: moves search to backend for better experience. It does remove show search results in context. 

amir20:


amir20:
- [ ]  Need to escape html

amir20:
Related to https://github.com/amir20/dozzle/issues/3240

PR #3256, opened 2024-09-05
fix: fixes search while looking back history

amir20:


amir20:
Yea, I noticed that. That's a UI issue. I only load older logs when I see at least one log. I think that's a bug. Not sure how to fix it yet. 🤷🏼‍♂️

Issue #3257, opened 2024-09-06
Agent crashing shortly after running


grosejon:
### 🔍 Check for existing issues

- [X] Completed

### How is Dozzle deployed?

Agents

### 📦 Dozzle version

8.4.1

### ✅ Command used to run Dozzle

services:
  dozzle-agent:
    image: amir20/dozzle:latest
    container_name: dozzle-agent-beelink
    command: agent
    networks:
      - socket_proxy
    healthcheck:
      test: ["CMD", "/dozzle", "healthcheck"]
      interval: 5s
      retries: 5
      start_period: 5s
      start_interval: 5s
    environment:
      - DOZZLE_HOSTNAME=beelink
      - DOCKER_HOST=tcp://socket-proxy:2375
    ports:
      - 7007:7007

networks:
  socket_proxy:
    external: true

### 🐛 Describe the bug / provide steps to reproduce it

After running for a short time, the agent will crash. I'm running the main instance on a synology, and then two agents: one on another synology (working), and the other on a beelink mini pc running ubuntu (error). The two agents are using the same compose, but the beelink one keeps crashing. 

### 💻 Environment

Client: Docker Engine - Community
 Version:    27.2.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 6
  Running: 5
  Paused: 0
  Stopped: 1
 Images: 10
 Server Version: 27.2.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 472731909fa34bd7bc9c087e4c27943f9835f111
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
[...]

PR #3259, opened 2024-09-06
chore: adds fatal error message when event.message is nil

amir20:


amir20:
See https://github.com/amir20/dozzle/issues/3257

PR #3261, opened 2024-09-07
fix: improves search by backfilling on first search

amir20:


amir20:
@FrancYescO I think this might fix search. 

Question for you. I have been struggling with something and can't figure out which is the right way. Do you think searching json logs should only search visible keys? Or that it should search everything. 

For example, log: `{a: "foo", b:"bar"}`. Say for example I hid b and only showing a. So essentially Dozzle would show `a=foo`. If user searches "bar", should that log be shown?

FrancYescO:
I think should do both (in this way you have the possibility to search in a value of a specific key prepending the key to the value)...but if you are asking maybe the best choice is to put a checkbox "search only in json value"

Ps. I also got some strange bug in the past where the order of a json log was very important, so if you are parsing a json check that is not altering the order (at least if you are not using the sorting feature):
{a: "foo", b:"bar"}
is not the same
{b:"bar", a: "foo"}

FrancYescO:
Oh yeah, searching only the visible can be enough

PR #3293, opened 2024-09-26
feat: uses bcrypt hash instead

amir20:
See https://github.com/amir20/dozzle/security/advisories/GHSA-w7qr-q9fh-fj35


mohammed90:
Sorry I was busy with other things and couldn't reply to your comments on the report. Thank you for the fix!

PR #3294, opened 2024-09-28
fix: guess log level will also check severity

EliasAx:
It's fairly common to have severity as the key for the log level. For example google cloud uses it. So this PR adds a check to check for the severity key if the level key is not found.


amir20:
❤️

amir20:
Releasing `8.5.4` to include this.

Issue #3298, opened 2024-09-30
Log does not always render properly


aetherpw:
### 🔍 Check for existing issues

- [X] Completed

### How is Dozzle deployed?

Standalone Deployment

### 📦 Dozzle version

8.5.3

### ✅ Command used to run Dozzle

```yml
 dozzle:
    container_name: dozzle
    image: amir20/dozzle:latest
    restart: unless-stopped
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      DOZZLE_ENABLE_ACTIONS: true
      DOZZLE_LEVEL: debug
      DOZZLE_NO_ANALYTICS: true
      DOZZLE_AUTH_PROVIDER: forward-proxy
      DOZZLE_AUTH_HEADER_USER: Cf-Access-Authenticated-User-Email
      DOZZLE_AUTH_HEADER_EMAIL: Cf-Access-Authenticated-User-Email
      DOZZLE_AUTH_HEADER_NAME: Cf-Access-Authenticated-User-Email
    ports:
      - "8899:8080"
    healthcheck:
      test: ["CMD", "/dozzle", "healthcheck"]
      interval: 5s
      timeout: 30s
      retries: 5
      start_period: 30s
```

### 🐛 Describe the bug / provide steps to reproduce it

Dozzle does not properly render emitted logs by grafana (*not exclusive to grafana*)

### 💻 Environment

```yml 
Client: Docker Engine - Community
 Version:    27.2.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 9
  Running: 9
  Paused: 0
  Stopped: 0
 Images: 9
 Server Version: 27.2.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
[...]

PR #3300, opened 2024-09-30
fix: supports empty values for logfmt

amir20:


amir20:
Fixes https://github.com/amir20/dozzle/issues/3298

Issue #3221, opened 2024-08-21
Seeing logs for stopped containers


konradstrachan:
### Describe the feature you would like to see

Being able to sww stopped containers and access their logs in exactly the same way as running containers

### Describe how you would like to see this feature implemented

Within the list of containers show both running and stopped, potentially differentiated by the running containers using the usual text colour and the stopped containers using a lighter colour to distinguish between the two types of containers

### Describe any alternatives you've considered

I've read the FAQs and searched around but can't find any reference to how to make stopped containers show their logs using Dozzle. Portainer enables this, but it provides a lot of additional functionality I don't want to expose for this use case.

amir20:
@konradstrachan I am not sure what you are asking. Have you enabled showing stopped containers in settings? 

<img width="1063" alt="Screenshot 2024-08-21 at 9 13 15 AM" src="https://github.com/user-attachments/assets/ed106f86-9eb4-4d91-8496-8b888d971360">

konradstrachan:
Doh! I read through everything but somehow missed this option. As a suggestion it's worthwhile having this in the FAQ in case someone else misses it like me. Thanks for the reply

undaunt:
I agreed with @konradstrachan as I searched the docs website for this and couldn't locate it.

amir20:
PRs welcomed to update documentation. 🙋🏼

PR #3304, opened 2024-10-02
docs: Add Show Stopped Containers to Dozzle

ddxv:
Include explanation of where all containers option is as referenced in #3221. 

Is there a way to do this via `docker run` and `docker compose` so those can be added to the FAQ as well? I saw there is a setting `showAllContainers:true` but am not sure how it should be added.


amir20:
Thanks. I'll be merging this. 

> Is there a way to do this via docker run and docker compose so those can be added to the FAQ as well? I saw there is a setting showAllContainers:true but am not sure how it should be added.

Not that's not possible. Mainly because these UI configurations are in javascript so passing them from CLI to client is pretty hard. I would have to handle conflicts which can get complicated. 

The better option is to use [multi-user auth](https://dozzle.dev/guide/authentication) so that Dozzle can save profiles to disk. This ensures that when a user changes their settings, the updates will be reflected on all devices. Without authentication, Dozzle cannot store user settings because there is no concept of a user.

Issue #3309, opened 2024-10-04
Option to remember user (stay logged in)


FibreTTP:
### Describe the feature you would like to see

Related to #2806, which was closed after fixing an issue with redirect URLs, which was only one part of the feature request.

An option to persist user login state through browser sessions should be added. Length of validity could also be configurable with an environment variable.

### Describe how you would like to see this feature implemented

Set an exp claim in the JWT and set the same value in the cookie expires field? I assume that would work, but I don't know anything about Go or even JWTs in general so I can't really help here, sorry.

I think these are the relevant places in the code:
https://github.com/amir20/dozzle/blob/9fa4911aec72ae7b14eeade2b272f62e1b9a4969/internal/auth/simple.go#L39-L42
https://github.com/amir20/dozzle/blob/077eeb79170d3286cae8e817306a927f35b46ddf/internal/web/auth.go#L14-L21

And upstream:
https://github.com/go-chi/jwtauth/blob/c856efe2bae161976c2f8d9899f0ee1d1b77056e/jwtauth.go#L122-L135

FibreTTP:
I check logs to make sure things aren't broken (e.g., Gluetun) often enough that even having to open Bitwarden and simply clicking auto-fill is getting annoying. I don't always have a web browser open, and obviously session cookies are cleared when closed.

It would be nice if the duration is configurable, but I feel like one day is a good default.

amir20:
I see. This was easy enough solution that I implemented it in https://github.com/amir20/dozzle/pull/3313. 

I haven't added documentation yet, but you should be able to do `DOZZLE_AUTH_TTL=24h`. Only minutes and hours is supported right now. 

Give it a try and I can merge if it works. You should be able to test with `amir20/dozzle:pr-3313`.

amir20:
@FibreTTP good call out. Turns out I was not using expiration correctly with jwts. I fixed it now. I tried testing it by making the cookie expire longer and indeed it expires as expected. 

Can you give another try?

amir20:
BTW, there is a lot of changes in this release so you might notice some other changes.

amir20:
Merged now. Will need to do testing before releasing.

amir20:
Everything should be released now.

PR #3313, opened 2024-10-04
feat: supports auth ttl to override session with DOZZLE_AUTH_TTL env.

amir20:
Fixes https://github.com/amir20/dozzle/issues/3309

### DIFF
--- a/docs/guide/authentication.md
+++ b/docs/guide/authentication.md
@@ -62,6 +62,34 @@ users:
 
 Dozzle uses [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token) to generate tokens for authentication. This token is saved in a cookie.
 
+### Extending Authentication Cookie Lifetime
+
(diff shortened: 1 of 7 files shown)


FibreTTP:
This works well, but since there is no expiry claim set in the JWT, the Expiry of the cookie can be set to whatever and Dozzle will accept it (I've tested this). JWTs without an expiry claim never expire.

For my local-only usecase this is acceptable, but definitely not _good_.

I'm also not sure what the `timestamp` claim you're setting is doing. It seems the standard is to use `iat` (Issued At).

amir20:
That's a good point. I currently don't do anything with JWT expiration. But to do this right I probably should.

amir20:
JWT now properly expires. Can you test out again? Turns out I wasn't using expiration correctly.

FibreTTP:
Looks good and works :)

Issue #3285, opened 2024-09-22
List of exposed ports


pmarki:
### Describe the feature you would like to see

Hi, great work! I use it all the time, now when there is an option to start/stop containers I'd love to see ports mapped so this application could replace docker management and simply dashboards (I actually keep a dashboard only top keep list of ports being used in my docker setup). Have you thought about this maybe?

### Describe how you would like to see this feature implemented

List of containers on starting page shows also ports that are exposed

### Describe any alternatives you've considered

Plenty of dashboards provides this feature, often they are over complicated or labels have to be kept in sync with docker compose configuration.

amir20:
Hello,

Thank you! I have a couple of questions for you:

1. One principle I want to maintain for Dozzle is monitoring frequently changing data, such as stats, logs, and health checks. I can't imagine ports changing often. What is your use case? Do you need to check this frequently?
2. If it's easy to add, I’m not concerned about the principle. Where do you think it should be displayed? I believe the homepage dashboard is the most suitable location. The container log has a lot of UI elements, but it may not be used by many people.

Let me know! Thanks!

pmarki:
Hi,
Let me explain a bit more. 

At the moment we have a list of containers on starting page, there is a name, host, status, created and average metrics, there could be also a column for mapped ports. To avoid increasing width of the container status could be changed to an icon maybe (or even icon with drop down to stop/start the service?)

Another idea would be turn this table more into cards, similar to how hosts are now presented, sorting could be done as a dropdown next to "Rows per page", that could allow to have all information visible on mobile view and fixed 'kind of' an issue where cell text is truncated in narrow window.

My use case is: 
One place to manage containers: I can stop/start them, I can checks logs, I can see how I can access the service without looking into docker-compose.yaml to find out which port has been assigned to this container, I can quickly find available port when spinning up new container. Based on this I can say your project is close to perfection :)  I could try to play with the code and propose something myself but I thought worth asking first.

amir20:
Just for reference, I collect data and seems like on average the maximum number of containers are around 10k. 
<img width="941" alt="Screenshot 2024-09-26 at 8 11 34 AM" src="https://github.com/user-attachments/assets/69b3295e-2e5c-42fc-b8b2-80cc85a2794d">

I would consider this data when implementing UI changes.

pmarki:
Hi @amir20, 
Once again words of appreciation to you (and all the people sharing their awesome work for free).
I will try to implement this. To be honest I already managed to have something.

[...]

PR #3316, opened 2024-10-08
feat: Add ports

pmarki:
Related to [#3285](https://github.com/amir20/dozzle/issues/3285)

- Added 'ports' column to containers table on the dashboard
- Don't allow to sort by ports or details (as it doesn't make much sense)
- Added config to show/hide columns
- Added new column with popup with container details
- When status column is hidden - status will be displayed as icons next to container name
- Show all containers on the dashboard if "Show stopped containers" is checked
- Polish translations


pmarki:
hi @amir20 as promised though I probably have to ask for your help with checks, neither go nor TS are my languages, I appreciate any hint from you how to fix them. If you don't have time for this please let me know and I will dig deeper

amir20:
I'll take a look today 🙏

internal/agent/pb/rpc.pb.go
```diff
@@ -1,15 +1,15 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.34.2
-// 	protoc        v5.27.3
+// 	protoc        v3.12.4
```
amir20, line 4: This doesn't seem right. Two versions down graded

assets/pages/settings.vue
```diff
@@ -120,6 +120,30 @@
       </section>
     </section>
 
+    <section class="flex flex-col gap-2">
+      <div class="has-underline">
```
amir20, line 124: Do people really want all these settings?

amir20: If I merge I might trim these. This is too much customization IMO

internal/agent/pb/rpc.pb.go
```diff
@@ -315,10 +315,10 @@ type LogsBetweenDatesRequest struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	ContainerId string                 `protobuf:"bytes,1,opt,name=containerId,proto3" json:"containerId,omitempty"`
```
amir20, line 318: I am confused here, why are these types being changed?

amir20 (commented):
This is a very big PR. Much bigger than I had expected so it will take me some time to review to make sure nothing is broken. I am not sure if I agree with everything, I'll try to pull it locally and test it first to see what it looks like. 

Thanks for everything though.

amir20:
Looks like there is nil pointer in the test cases too

pmarki:
Hi @amir20 thank you for your time and I'm sorry to take it more then you expected.
My reasoning with all the changes was as follow:
- ports can take a bit of horizontal space when there are couple of them exposed in the container, someone who'd like to see them should make a decision and remove other column(s), so let's add a configuration to show/hide other columns but at the same time let's make all details still available (hence last column with popover). I didn't want to push only new column with exposed ports as it looks weird when most of it is truncated (and what is more important the truncated last digit is the one that matters). 

When it comes to downgrading: that's what I get on ubuntu 22.04, I had to install  protobuf and I had version 3.12. As mentioned I'm not familiar with go tooling but I can try to find the version 5.12 (probably all the auto generated changes comes from the version difference).

I can split the PR into smaller pieces, in the meantime I will try to fix a case you got (it worked for me but I might need go into other cases)

amir20:
> ports can take a bit of horizontal space when there are couple of them exposed in the container, someone who'd like to see them should make a decision and remove other column(s), so let's add a configuration to show/hide other columns but at the same time let's make all details still available (hence last column with popover). I didn't want to push only new column with exposed ports as it looks weird when most of it is truncated (and what is more important the truncated last digit is the one that matters).

This feels like it should be a drop down on the table where users can check or uncheck which columns they want. Settings page should be for global configurations. The table with different columns is only shown on the home page which makes more sense to put it there. Just in case you are not wasting too much, let's break these PRs to smaller ones so that we can work together before doing too much work. I can see your heart is in the right place and I actually think being able to show different columns is a good idea. 

> When it comes to downgrading: that's what I get on ubuntu 22.04, I had to install protobuf and I had version 3.12. As mentioned I'm not familiar with go tooling but I can try to find the version 5.12 (probably all the auto generated changes comes from the version difference).

It looks like you installed `github.com/golang/protobuf` which as an older version. You shouldn't need to install anything. You should be able to use `make tools` and it should auto install everything needed. I wouldn't expect for this PR for `go.mod` to change. 

> I can split the PR into smaller pieces, in the meantime I will try to fix a case you got (it worked for me but I might need go into other cases)

Great. I would first send one PR with just the port mappings to match `docker ps`. This is so that it reduces confusion. If someone uses `docker ps` but then comes back to Dozzle and sees something different, it would be confusing. In this PR, let's not worry about settings to show/hide columns. It should just be the agent, agent server, docker client and UI. That's probably already big enough. Also don't forget there is two places you need to read the port  `newContainerFromJSON` and `newContainer`.  `newContainerFromJSON` is used for running containers and `newContainer` is used when getting a list.

pmarki:
hi @amir20 
I went through port mapping and I think it works correctly I'm just using wrong terminology: I wanted to show a list of published ports (the one that can be used from outside), of course those are ports that have to be unique and that's where my use case is from - when adding new container I'd like to have a way to quickly see which ports are occupied or when I rarely  use a service to find out which port should I add to url to access web service. Docker ps command shows both exposed and published ports, that's why it's confusing. 

`            "Ports": {
                "9200/tcp": [
                    {
                        "HostIp": "0.0.0.0",
                        "HostPort": "9201"
                    },
                    {
                        "HostIp": "::",
                        "HostPort": "9201"
                    }
                ],
                "9300/tcp": null,
                "9600/tcp": [
                    {
                        "HostIp": "0.0.0.0",
                        "HostPort": "9601"
                    },
                    {
                        "HostIp": "::",
                        "HostPort": "9601"
                    }
                ],
                "9650/tcp": null
            },
`
from this container it will only show 6201 and 9601.

I'd like to keep those two types of ports separated, but of course that's your project. Maybe I can go with popup from the last column for now with two new fields: 'exposed ports' and 'published ports'? What's your view on this?

amir20:
Going to close this since it can't be merged in this state.

PR #3346, opened 2024-10-25
fix: limit default SQL log entries to 100

yubiuser:
When opening the SQL Analytics, it fetches all logs by default. On containers with many logs this can take a while. This PR changes the default by limiting it to first 100 log entries

### DIFF
--- a/assets/components/LogViewer/LogAnalytics.vue
+++ b/assets/components/LogViewer/LogAnalytics.vue
@@ -31,7 +31,7 @@
 import { Container } from "@/models/Container";
 import { type Table } from "@apache-arrow/esnext-esm";
 const { container } = defineProps<{ container: Container }>();
-const query = ref("SELECT * FROM logs");
+const query = ref("SELECT * FROM logs LIMIT 100");
 const error = ref<string | null>(null);
 const debouncedQuery = debouncedRef(query, 500);
 const evaluating = ref(false);


amir20:
I don't think this is the right implementation. There is currently a [limit of 1000](https://github.com/amir20/dozzle/blob/96fbee9ece931d22e2f58e0bb92d801ddc4f0ff3/assets/components/LogViewer/LogAnalytics.vue#L38).

I plan to add pagination so that it can be paginated. I tried doing virtual scroll, but that was just too complicated with tables.  

Going to close this. If you know Vue, and you can help out, it would be better to introduce a pagination of maybe 100 records. Similar to what I am doing on the homepage.

yubiuser:
Your limit of 1000 is not the relevant limit here because it happens after SQL processes all the data. On slower hardware with a lot of log entries this can still overwhelm the machine. 
See this CPU graph. The load is caused by the dozzle container opening SQL Analytics tool - without ever being able to retrieve the data - I always canceled it after a few minutes.

The host machine is a Intel Celeron J4105 - not fast but also not an RPi.

amir20:
That's a valid point. So this is only to make it faster initially.

amir20:
Going to merge this. But I guess if the user tries doing a SQL without `LIMIT`, then they know what they are doing.

PR #3352, opened 2024-10-27
fix: fixes base ansi-to-html colors

Jeto143:
This fixes a bug with base colors passed to https://github.com/rburns/ansi-to-html.

The `fg` and `lg` flag expect valid CSS, however `--base-color` and `--base-content-color` are LCH values, which can't directly be used inside CSS. For them to properly work, [`lch()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/lch) needs to be used.

One way to test this (which is also how I noticed it) is by using [chalk](https://github.com/chalk/chalk) and running the following statement. The whole text would end up red instead of only the part at the start, due to the base color failing to be applied to the rest of the string.

```js
console.log(`${chalk.red("This is red")} but this is not`);
```

### DIFF
--- a/assets/components/LogViewer/SimpleLogItem.vue
+++ b/assets/components/LogViewer/SimpleLogItem.vue
@@ -23,8 +23,8 @@ import stripAnsi from "strip-ansi";
 
 const ansiConvertor = new AnsiConvertor({
   escapeXML: false,
-  fg: "var(--base-content-color)",
-  bg: "var(--base-color)",
+  fg: "lch(var(--base-content-color))",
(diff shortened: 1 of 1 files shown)


amir20:
Thanks! I'll have to check this out locally because the colors are actually in `oklch` format. So the PR might be updated.

amir20:
Merged in https://github.com/amir20/dozzle/pull/3353. `oklch` was the right css to use. 

Screenshot 2024-10-27 at 1 38 55 PM

Thanks!

PR #3388, opened 2024-11-14
feat: show loader while opening event source

amir20:


amir20:
I have been playing with this. Now the connection loads sooo fast that the loader doesn't even show.

PR #3394, opened 2024-11-15
feat: enables swarm cluster to connect to external agents

amir20:


amir20:
I'll need to do more testing before merging and update the documentation.

amir20:
OK I did more testing. Looks good! I think my local swarm cluster is broken. But a new one worked.

PR #3357, opened 2024-10-28
fix: fix link color when using colorized logs

Jeto143:
Following up on https://github.com/amir20/dozzle/pull/3353, this PR fixes a minor bug where HTML links are not shown with their normal color when using ANSI color escape sequences due to https://github.com/rburns/ansi-to-html inserting HTML tags inside `.log-wrapper`.

With:

```js
console.log(`${chalk.red('hi')} https://www.google.com ${chalk.green('world')}`);
```

Before:

image

After:

image

PR #3440, opened 2024-12-08
fix: fix ansi link color + regex

Jeto143:
This fixes a regression of https://github.com/amir20/dozzle/pull/3357 which was due to the relevant style being moved over to another component.

Before:

image

After:

image

This also solidifies the regex for matching URLs, as it could cause issues in some cases such as when an URL is at the very end of a log entry.

Before:

image

After:

image

### DIFF
--- a/assets/components/LogViewer/SimpleLogItem.vue
+++ b/assets/components/LogViewer/SimpleLogItem.vue
@@ -28,7 +28,8 @@ defineProps<{
 }>();
 
 const colorize = (value: string) => ansiConvertor.toHtml(value);
-const urlPattern = /(https?:\/\/[^\s]+)/g;
+const urlPattern =
+  /https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b[-a-zA-Z0-9()@:%_+.~#?&\/=]*/g;
 const linkify = (text: string) =>
   text.replace(urlPattern, (url) => `<a href="${url}" target="_blank" rel="noopener noreferrer">${url}</a>`);
 </script>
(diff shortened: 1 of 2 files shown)


assets/components/LogViewer/SimpleLogItem.vue
```diff
@@ -28,7 +28,14 @@ defineProps<{
 }>();
 
 const colorize = (value: string) => ansiConvertor.toHtml(value);
-const urlPattern = /(https?:\/\/[^\s]+)/g;
+const urlPattern =
+  /https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b[-a-zA-Z0-9()@:%_+.~#?&\/=]*/g;
```
Jeto143, line 32: This is based on [this](https://stackoverflow.com/a/3809435/965834) and slightly improved (Stack Overflow isn't always a reliable source but when an answer has been maintainted/upvoted for so long, it's probably a good sign).

amir20:
Thanks. Will review!

Jeto143:
> Thanks. Will review!

Thanks! I'll look into the failing tests (not sure what's up at first glance).

assets/components/LogViewer/SimpleLogItem.vue
amir20: Nice find.

amir20:
Looks good overall. I can merge when ready.

Jeto143:
> Looks like it's an easy fix. Some of the snapshots changed because there were changes to styles. You just need to `pnpm run test -u`.

Ah thanks, not super familiar with Jest, I was trying to fix it manually haha. Should be fine now (let's see)!

Issue #3478, opened 2024-12-23
Traefik log levels in traefik v3.x are no longer automatically recognized or filterable


arudnitsky:
### 🔍 Check for existing issues

- [X] Completed

### How is Dozzle deployed?

Standalone Deployment

### 📦 Dozzle version

v8.9.0

### ✅ Command used to run Dozzle

docker run -d --restart=unless-stopped --name dozzle --volume=/var/run/docker.sock:/var/run/docker.sock -p 0.0.0.0:7070:7070 --env DOZZLE_ADDR=:7070 amir20/dozzle:latest

### 🐛 Describe the bug / provide steps to reproduce it

I upgraded traefik from v2.x to v3.1 and Dozzle can no longer filter by log levels. In traefik v3, log levels are now three characters: DBG, INF, WRN, etc... whereas in v2, they were given by DEBUG, INFO, WARNING.

Everthing messages now found under Unknown. If the word "error" is detected in a message, a red dot is placed in the UI before the message.

Example traefik v3 logs:
```
2024-12-23T16:51:00Z INF github.com/traefik/traefik/v3/cmd/traefik/traefik.go:611 > Stats collection is enabled.

2024-12-23T16:51:00Z DBG github.com/traefik/traefik/v3/pkg/server/server_entrypoint_tcp.go:231 > Starting TCP Server entryPointName=http

2024-12-23T17:01:30Z WRN github.com/traefik/traefik/v3/pkg/version/version.go:76 > Error checking new version error="Get \"https://update.traefik.io/repos/traefik/traefik/releases\": dial tcp 13.39.208.199:443: i/o timeout"
```

### 💻 Environment

Client: Docker Engine - Community
 Version:    27.4.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.31.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 4
  Running: 4
  Paused: 0
  Stopped: 0
 Images: 10
 Server Version: 27.4.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
[...]

PR #3491, opened 2024-12-30
fix: fixes log levels not being parsed correctly and supports short levels like dbg

amir20:
Fixes https://github.com/amir20/dozzle/issues/3478


arudnitsky:
Thank you!

PR #3550, opened 2025-01-14
feat: add more tags to same release

### LINKED ISSUE #3549: Support minor release for ghcr.io

Several projects support adding major.minor in semantic mode, for example:

This works: ghcr.io/amir20/dozzle:v8.10.1
This not works: ghcr.io/amir20/dozzle:v8.10
This not works: ghcr.io/amir20/dozzle:v8

Add the tag for the minor and if make sense for the major.
Will be nice to update automatic when minor changes happens for the agent without being forced to use the latest.

_No response_

webysther:
I don't know if https://github.com/docker/metadata-action?tab=readme-ov-file#tags-input fail with 'v' prefix, I think not, but I used the 'raw' type because the result is the same after all and can be used w/o risk to changes on semver in future.

Closes #3549

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -111,6 +111,11 @@ jobs:
           images: |
             amir20/dozzle
             ghcr.io/amir20/dozzle
+          tags: |
+            type=raw,pattern=v{{version}}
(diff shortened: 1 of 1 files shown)


amir20:
I'll merge. No way of knowing until I release though.

Issue #3630, opened 2025-02-14
Support namespaces in k8s


amir20:
Hello,

I've just had a quick try on my k8s cluster. It's a success, the logs are up correctly!
I noticed that there was only the namespace default whitelist, is this a deliberate choice?

Thanks for the support ;)

_Originally posted by @JeuFore in https://github.com/amir20/dozzle/discussions/3614#discussioncomment-12203246_

PR #3631, opened 2025-02-14
feat: supports namespaces properly in k8s mode

amir20:
Fixes https://github.com/amir20/dozzle/issues/3630

### DIFF
--- a/internal/container/container_store.go
+++ b/internal/container/container_store.go
@@ -84,7 +84,7 @@ func (s *ContainerStore) checkConnectivity() error {
 			}
 
 			running := lo.Filter(containers, func(item Container, index int) bool {
-				return item.State != "exited"
+				return item.State != "exited" && item.FullyLoaded == false
 			})
 
 			sem := semaphore.NewWeighted(maxFetchParallelism)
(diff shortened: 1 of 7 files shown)


JeuFore:
I've just tested on all the namespaces in my cluster and the application crashes in a loop every time after 2m.

A memory address problem apparently.
I don't limit the application in resource

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9a9c19]

goroutine 14 [running]:
github.com/amir20/dozzle/internal/utils.(*RingBuffer[...]).Push(0x25e7420, {{0x0, 0x0}, 0x3ff0000000000000, 0x0, 0x41bfe0a000000000})
	/dozzle/internal/utils/ring_buffer.go:38 +0x59
github.com/amir20/dozzle/internal/container.(*ContainerStore).init(0xc0002b23f0)
	/dozzle/internal/container/container_store.go:397 +0x232
created by github.com/amir20/dozzle/internal/container.NewContainerStore in goroutine 1
	/dozzle/internal/container/container_store.go:54 +0x2db
```

amir20:
I was able to reproduce it. It also turns out that when fetching a specific pod, I have to pass the pod.namespace exactly. 

Try the latest version. With my testing it works fine.

JeuFore:
Great, the problem's been fixed.
Everything is retrieved quickly, logs and metrics, and now I have 16 pages of containers 👍

amir20:
Awesome. Going to merge.

amir20:
@JeuFore That's a lot of containers 😂 But I have seen up to thousands. Performance wise it's ok. I think the UI could be better.

PR #3635, opened 2025-02-16
fix: fixes ansi-to-html fallback colors

Jeto143:
This fixes wrong CSS variables being used for fallback colors, as daisyUI 5 changed them.

Note: no need for `oklch` anymore since they're now already in that format in main.css.

With the following code (using [chalk](https://github.com/chalk/chalk)):

```ts
console.log(chalk.bgRed('Hello') + ' World');
console.log(chalk.red('Hello') + ' World');
```

Before:

image
image

After:

image
image


amir20:
Thanks!

PR #3648, opened 2025-02-19
feat: Display actions dropdown on mobile

Crashillo:
Displays the actions dropdown in the mobile menu, and rearranges the toolbar items in order to fit.

According what proposed here: https://github.com/amir20/dozzle/commit/1637b37cc15e24e23036e78b7e8991a61dc8b998#commitcomment-152751497

### DIFF
--- a/assets/components/GroupedViewer/GroupedLog.vue
+++ b/assets/components/GroupedViewer/GroupedLog.vue
@@ -1,14 +1,14 @@
 <template>
   <ScrollableView :scrollable="scrollable" v-if="group.containers.length && ready">
     <template #header>
-      <div class="mx-2 flex items-center gap-2 md:ml-4">
+      <div class="mx-2 flex flex-wrap items-center gap-2 md:ml-4">
         <div class="@container flex flex-1 gap-1.5 truncate md:gap-2">
           <div class="inline-flex font-mono text-sm">
             <div class="font-semibold">{{ $t("label.container", group.containers.length) }}</div>
           </div>
         </div>
-        <MultiContainerStat class="ml-auto" :containers="group.containers" />
-        <MultiContainerActionToolbar class="max-md:hidden" @clear="viewer?.clear()" />
+        <MultiContainerStat class="ml-auto justify-center max-md:basis-full max-md:order-last" :containers="group.containers" />
+        <MultiContainerActionToolbar @clear="viewer?.clear()" />
       </div>
     </template>
     <template #default>
--- a/assets/components/HostViewer/HostLog.vue
+++ b/assets/components/HostViewer/HostLog.vue
@@ -1,7 +1,7 @@
 <template>
   <ScrollableView :scrollable="scrollable" v-if="host">
     <template #header>
-      <div class="mx-2 flex items-center gap-2 md:ml-4">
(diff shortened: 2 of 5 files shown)


amir20:
You missed a bunch of instances 

```
❯ rg MultiContainerStat
assets/components/HostViewer/HostLog.vue
14:        <MultiContainerStat class="ml-auto" :containers="containers" />

assets/components/GroupedViewer/GroupedLog.vue
10:        <MultiContainerStat class="ml-auto justify-center max-md:basis-full max-md:order-last" :containers="group.containers" />

assets/components/StackViewer/StackLog.vue
17:        <MultiContainerStat class="ml-auto" :containers="stack.containers" />

assets/components/MultiContainerViewer/MultiContainerLog.vue
11:        <MultiContainerStat class="ml-auto" :containers="containers" />

assets/components/ContainerViewer/ContainerLog.vue
6:        <MultiContainerStat

assets/components/ServiceViewer/ServiceLog.vue
14:        <MultiContainerStat class="ml-auto" :containers="service.containers" />
```

amir20 (changes requested):
This is only implemented in one spot. See comment on all instances that need to be changed.

Crashillo:
`ContainerLog` looks to have an slightly different markup, so I presume it's a different context. I cannot find where is rendered.

amir20:
`ContainerLog` is actually the main view when rendering single containers. The ones you implemented are the services, custom groups, stacks, etc....

Have you tested this locally to make sure it works? Because I imagine `ContainerLog` would be the easiest to test.

Crashillo:
I could not make it work. After installing the resources, and run the steps you mention, `make dev` always returned an error that I couldn't figure it out. Since these changes only involved some tailwind styles, I didn't like to get complicated.

amir20:
Hmmm I tested this. This is really ugly. 


Screenshot 2025-02-19 at 12 12 33 PM

It takes a lot of space. I would personally not like it. I am not sure how I would feel about merging it. 

I have a lot of users on mobile who appreciate the space. So if you can spend time making it better then I can merge. Otherwise, let's not do.

amir20:
> I could not make it work. After installing the resources, and run the steps you mention, make dev always returned an error that I couldn't figure it out. Since these changes only involved some tailwind styles, I didn't like to get complicated.

You'll probably want to run it locally so you can test it.

amir20:
Closing for now.

Crashillo:
IMHO, it's far more useful a menu than displaying some stats, but it's your application. Don't like it? Ok, up to you. It was just a suggestion.

amir20:
> IMHO, it's far more useful a menu than displaying some stats, but it's your application. Don't like it? Ok, up to you. It was just a suggestion.

You just have to put time in to it and make it work. If you can't get it to work locally then I don't know what to tell you. I am not going to merge changes blindly as I'll have to test it myself. This PR doesn't even work because it doesn't correctly implement it for a single container. 

There is a lot of users on Dozzle and at least a few have asked for stats on mobile. It's really hard to make all users happy. So I think a good compromise would be making the menu a small dot or something but still showing everything in one line. That's probably what I would do if I had time.

PR #3656, opened 2025-02-21
feat: Supports log levels with common separators

garikkh:
Revival of [this PR](https://github.com/amir20/dozzle/pull/3368) that added support for log levels followed by an equal sign, but also added support for `/` and `-`. Original PR was removed in https://github.com/amir20/dozzle/pull/3491

`/` is especially a common format for celery:

```
[2025-02-20 14:55:36,789: INFO/MainProcess] Task example.tasks.add[1234abcd] succeeded in 0.015s: 42
[2025-02-20 14:56:10,123: ERROR/MainProcess] Task example.tasks.divide[5678efgh] raised exception: ZeroDivisionError('division by zero')
```

### DIFF
--- a/internal/container/level_guesser.go
+++ b/internal/container/level_guesser.go
@@ -22,6 +22,7 @@ var logLevels = [][]string{
 
 var plainLevels = map[string][]*regexp.Regexp{}
 var bracketLevels = map[string][]*regexp.Regexp{}
+var separatorLevels = map[string][]*regexp.Regexp{}
 var timestampRegex = regexp.MustCompile(`^(?:\d{4}[-/]\d{2}[-/]\d{2}(?:[T ](?:\d{2}:\d{2}:\d{2}(?:\.\d+)?Z?|\d{2}:\d{2}(?:AM|PM)))?\s+)`)
 
 func init() {
@@ -36,6 +37,7 @@ func init() {
 		first := levelGroup[0]
 		for _, level := range levelGroup {
 			bracketLevels[first] = append(bracketLevels[first], regexp.MustCompile("(?i)\\[ ?"+level+" ?\\]"))
+			separatorLevels[first] = append(separatorLevels[first], regexp.MustCompile("(?i)"+level+"[=/-]"))
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


amir20:
I am trying to figure out why I removed `error=` logs. If I remember correctly, it was causing false positives. An example was `DBG blah blah error=foobar` should be debug and not error. 

I am ok with merging `LEVEL/` but I think `LEVEL=` will cause unwanted behavior.

garikkh:
@amir20 thanks for looking at this

I can remove the `=` case, or work on a different solution that will ensure `DBG blah blah error=foobar` works properly.

I was also considering making some changes to the guesser, do you have any objections to combining the regular expressions to using for example `(?i)^(?:error|err)[^a-z]` instead of two different regular expressions/match calls?

amir20:
> I was also considering making some changes to the guesser, do you have any objections to combining the regular expressions to using for example (?i)^(?:error|err)[^a-z] instead of two different regular expressions/match calls?

🎉 We should def do that. Regex is expensive. 

> work on a different solution that will ensure DBG blah blah error=foobar works properly

It would be good to do that. The problem is that I can't imagine how it would be possible. The presence of `error=` doesn't mean it is level ERROR. 

I am all for you trying to find a better logic to improve the guesser. I haven't spent a lot of time on it.  There is lots of tests so as long as they pass I think it will be ok.

garikkh:
> 🎉 We should def do that. Regex is expensive.

Yeah, and adding a third group of them was worrisome. Pushed my changes for it, but might come back and make some more changes.

> It would be good to do that. The problem is that I can't imagine how it would be possible. The presence of error= doesn't mean it is level ERROR.

But would marking this as ERROR be better than keeping it as unknown? It could be done after the entire loop, where just before it returns unknown, it will do a final check of `<level>=` messages

amir20:
> But would marking this as ERROR be better than keeping it as unknown? It could be done after the entire loop, where just before it returns unknown, it will do a final check of <level>= messages

That's a good question. I am not familiar with all logging libs out there.  I would say be conservative and don't assume `error=` means error for now.

amir20:
Let me know when the PR is ready for review.

garikkh:
ready for review now - I kept `-` as a separator too because I vaguely remember some node processes using that, but not super common so we can remove that if needed.

amir20:
Ok merged. Wait till `master` builds and give it a try.

amir20:
I'll release some time this weekend 🎉

Issue #3688, opened 2025-03-08
Add PWA Manifest


stasadance:
### Describe the feature you would like to see

I would love to see a proper PWA manifest in Dozzle frontend

### Describe how you would like to see this feature implemented

I mainly use Dozzle on mobile to monitor systems remotely. 

I have it added to my mobile home screen but every time I open it that way, it leads me to a new tab in the browser. 

Having a PWA would allow it to open similar to a real app, where it remembers the last thing I was looking at until I fully close it.

### Describe any alternatives you've considered

N/A

amir20:
Thanks for the suggestion @stasadance. 

Can you help me understand what feature of PWA you like? 

I think with [Vite PWA](https://vite-pwa-org.netlify.app/) it would be pretty easy to implement this.  But last time I tried it, I was not able to change the path dynamically. So `--base` wouldn't work.

stasadance:
@amir20 thanks for quick response!

I believe that just the `display: "standalone"` is the most important part I am looking for.

In my projects I just use Vite PWA with this kind of config and nothing more:

```ts
VitePWA({
    registerType: "autoUpdate",
    manifest: {
        display: "standalone",
        background_color: "#000000",
        theme_color: "#000000"
    }
})
```

amir20:
Ok, I tried to do this. According to Chrome, it has been installed correctly. Can you give it a try? You can pull `amir20/dozzle:pr-3691`.

I am not too familiar with what it _should_ do so perhaps you can test it out and let me know how it works. The PR is pretty simple.

PR #3691, opened 2025-03-08
feat: adds pwa in production mode

amir20:
Fixes https://github.com/amir20/dozzle/issues/3688

### DIFF
--- a/assets/main.ts
+++ b/assets/main.ts
@@ -7,3 +7,9 @@ Object.values(import.meta.glob<{ install: (app: VueApp) => void }>("./modules/*.
   i.install?.(app),
 );
 app.mount("#app");
+
+if ("serviceWorker" in navigator && import.meta.env.MODE === "production") {
+  navigator.serviceWorker.register(withBase("/sw.js"), {
+    scope: withBase("/"),
+  });
+}
--- a/assets/main.ts
+++ b/assets/main.ts
@@ -7,9 +7,3 @@ Object.values(import.meta.glob<{ install: (app: VueApp) => void }>("./modules/*.
   i.install?.(app),
 );
 app.mount("#app");
-
-if ("serviceWorker" in navigator && import.meta.env.MODE === "production") {
-  navigator.serviceWorker.register(withBase("/sw.js"), {
-    scope: withBase("/"),
-  });
-}
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "simple-git-hooks": "^2.11.1",
     "ts-node": "^10.9.2",
     "typescript": "^5.8.2",
+    "vite-plugin-pwa": "^0.21.1",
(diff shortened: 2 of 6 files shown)


stasadance:
Did a pull but I don't see a `<link rel="manifest" href="manifest.json" />` or similar file and it doesn't seem to act as a PWA

amir20:
I only installed the service worker because the URLs can change. I wasn't aware of a manifest.json. Let me see if that's something I can do.

amir20:
Ok that was a little more tricky. I added a manifest now. Wait till push is done and try again...

amir20:
@stasadance hmmmm it seems like the service worker breaks auth as the assets are cached. Do you need service worker? Can I do _JUST_ manifest?

stasadance:
> @stasadance hmmmm it seems like the service worker breaks auth as the assets are cached. Do you need service worker? Can I do _JUST_ manifest?

Manifest alone is fine.

stasadance:
Manifest works pretty well on my device. Adding the website to my home screen in iOS safari now results in an app like feel where it opens Dozzle in full screen instead of in the browser.

Small thing I noticed.. The site doesn't adjust for the phone's insets in standalone display mode so it looks a bit weird. Screenshot for reference.



I think this can be fixed with 2 changes:

- Something like `<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, viewport-fit=cover" />`
- Using CSS safe area variables in the style for some padding on mobile. [safe-area-inset-bottom](https://developer.mozilla.org/en-US/docs/Web/CSS/env#safe-area-inset-top)

Let me know if you would like me to work on these, I'm not familiar with local dev yet but I can set it up.

amir20:
That would be great if you can. I am not too familiar with PWA. You should be able to do `make tools` and finally `make dev`. 

I think Vite PWA does a lot more than needed. Perhaps just creating a manifest is good enough. Thoughts? 

Let me know if you have a hard time getting local dev to work. If you are on a Mac (which is what I have) it should be ok.

stasadance:
> That would be great if you can. I am not too familiar with PWA. You should be able to do `make tools` and finally `make dev`.
> 
> I think Vite PWA does a lot more than needed. Perhaps just creating a manifest is good enough. Thoughts?
> 
> Let me know if you have a hard time getting local dev to work. If you are on a Mac (which is what I have) it should be ok.

Manifest alone sounds good enough to me. I'll work on a duplicate PR with some CSS.

amir20:
I pushed some changes to remove vite pwa but keep manifest. 

I have never heard of `[safe-area-inset-bottom](https://developer.mozilla.org/en-US/docs/Web/CSS/env#safe-area-inset-top)` 😂

I'll wait to see what you cook up.

amir20:
Going to merge this and release since there are some bugs pending.

PR #3693, opened 2025-03-09
feat: safe area styling for ios pwa

stasadance:
This PR adds required meta tags and some safe area styling for PWA to render correctly on iOS.

I was able to test this by opening iOS Simulator in XCode and going to `http:/localhost:3100` in safari, and then pressing share button and saving it to home screen.

Before:

</img>

After:

</img>

### DIFF
--- a/assets/main.css
+++ b/assets/main.css
@@ -126,3 +126,12 @@
     }
   }
 }
+
+nav {
+  padding-top: env(safe-area-inset-top);
+}
+
+body {
+  padding-top: env(safe-area-inset-top);
+  padding-bottom: env(safe-area-inset-bottom);
+}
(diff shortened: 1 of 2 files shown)


amir20:
Thanks! I'll check it out tomorrow. 

I also have an iPhone but it doesn't do this. Is this a specific model?

stasadance:
> Thanks! I'll check it out tomorrow.
> 
> I also have an iPhone but it doesn't do this. Is this a specific model?

Depends if you have the notch or not, XCode iOS sim has all of them I think I tested with iPhone 15

amir20:
I am going to merge this. But I see no difference on my phone and simulator. I am using iPhone 16 and it is below the nudge. 

Screenshot 2025-03-09 at 8 29 32 AM


I think what would be nice is to support the nudge on the left. 
Screenshot 2025-03-09 at 8 30 50 AM

PR #3694, opened 2025-03-09
feat: adds inset for left and right

amir20:


amir20:
Related to https://github.com/amir20/dozzle/pull/3693

PR #3800, opened 2025-04-13
feat: Update Bottom Sidebar Button Border Color to #e6bd40

PiyushDixit96:
When I first tried out Dozzle, I had trouble locating the sidebar in light mode because it wasn’t clearly visible. To improve visibility, I updated the Dozzle icon color so that it stands out in both light and dark modes.

**If the admin finds this helpful, feel free to merge it — otherwise, no worries. Thanks!**

## Before
dark dark
light light

## After
yellow dark
yellow light


PiyushDixit96:
i am not good in css,

amir20:
You are right. On light theme, it's really hard to see it. 

Some thoughts: 

1. You should fix your commits so that not all of those commits show up as part of the PR. 
Screenshot 2025-04-13 at 8 57 58 AM

2. Using `style=` is not right. I use tailwind. You can probably how it's done in other places. 
3. Finally, I think the yellow color looks out of place.

amir20:
I play with this to see if I can find a better color.

amir20:
See https://github.com/amir20/dozzle/pull/3801

PiyushDixit96:
that's good

PiyushDixit96:
next time i will fix pr commit

Issue #3523, opened 2025-01-08
connecting to agent error: too many colons


fama-lama:
### 🔍 Check for existing issues

- [X] Completed

### How is Dozzle deployed?

Remote Hosts

### 📦 Dozzle version

8.9.1

### ✅ Command used to run Dozzle

```

  dozzle:
    image: amir20/dozzle:latest
    container_name: dozzle
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - DOZZLE_REMOTE_AGENT=tcp://192.168.100.52:7007,tcp://192.168.100.53:7007,tcp://192.168.100.54:7007,tcp://192.168.100.56:7007,tcp://192.168.100.57:7007,tcp://192.168.100.59:7007,tcp://192.168.100.60:7007
    ports:
      - 8090:8080
```

### 🐛 Describe the bug / provide steps to reproduce it

deployed agents in all my lxcs. when trying to connect from my main dozzle instance, all the agents show offline.
logs seems to have one thing in common.

> "error info: address tcp://xx:xx:xx:xx:7007:443: too many colons in address"

port 443 is appended to all my addresses, eventhough i have already declared a port. 

### 💻 Environment

```
root@home-stack:~# docker system info
Client: Docker Engine - Community
 Version:    24.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.20.3
    Path:     /root/.docker/cli-plugins/docker-compose

Server:
 Containers: 12
  Running: 12
  Paused: 0
  Stopped: 0
 Images: 16
 Server Version: 24.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: journald
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8165feabfdfe38c65b599c4993d227328c231fca
[...]

PR #3893, opened 2025-05-16
docs: remove leading tcp://

adderek:
Examples were wrong, see
[#3523](https://github.com/amir20/dozzle/issues/3523) where tcp://1.2.3.4:7007 gets transformed into tcp://1.2.3.4:7007:443

Root cause: the endpoint values are passed to google golang grpc library, NewClient, which does not support "tcp://" prefix like other parts. Instead it supports
none, "dns://", "zookeeper://", "unix://", "unix-abstract://", "vsock:", "ipv4:", "ipv6:". In the future additional syntax might be added

See https://pkg.go.dev/google.golang.org/grpc#NewClient for GRPC documentation
and https://github.com/grpc/grpc/blob/master/doc/naming.md for possible values

### DIFF
--- a/docs/guide/supported-env-vars.md
+++ b/docs/guide/supported-env-vars.md
@@ -27,7 +27,7 @@ Configurations can be done with flags or environment variables. The table below 
 | `--namespace`         | `DOZZLE_NAMESPACE`         | `""`           |
 
 > [!TIP]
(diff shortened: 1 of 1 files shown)


amir20:
Good catch. Remote agents should not have `tcp`.

PR #3940, opened 2025-06-01
fix: Adjusted `<meter>` thresholds for CPU and Memory usage display

homelab-alpha:
## Description

This PR updates the `low` and `high` attributes of the HTML5 `<meter>` elements in `ContainerTable.vue`.

### What changed?

- Adjusted the visual scaling of CPU and Memory meters to be less aggressive.
- Updated thresholds:

  - **CPU**: `low=60`, `high=80` (was 40/70)
  - **Memory**: `low=70`, `high=90` (was 40/70)

### Why?

- Previous values triggered color shifts too early, even under normal load.
- These changes better reflect actual performance ranges and enhance visual clarity.

### Impact

- Purely UI-level change.
- No functional or data logic altered.

### Discussions

- Relates to #3939

### DIFF
--- a/assets/components/ContainerTable.vue
+++ b/assets/components/ContainerTable.vue
@@ -83,9 +83,9 @@
                   class="flex-1 overflow-hidden rounded-3xl"
                   min="0"
                   max="100"
-                  low="40"
+                  low="60"
                   optimum="0"
-                  high="70"
+                  high="80"
                   :value="Math.min(container.movingAverage.cpu, 100)"
                 ></meter>
                 <span class="w-8 text-right text-sm">{{ container.movingAverage.cpu.toFixed(0) }}%</span>
@@ -97,9 +97,9 @@
                   class="flex-1 overflow-hidden rounded-3xl"
                   min="0"
                   max="100"
-                  low="40"
+                  low="70"
                   optimum="0"
-                  high="70"
+                  high="90"
                   :value="container.movingAverage.memory"
                 ></meter>
                 <span class="w-8 text-right text-sm">{{ container.movingAverage.memory.toFixed(0) }}%</span>


homelab-alpha:
I tested it, and this is the result.

- **UI Modifications**: Highlight any changes made to the user interface.
- **Before & After**: Include screenshots or comparisons (if applicable).

| Event              | Before                | After                |
| ------------------ | --------------------- | -------------------- |
| HTML5 meter        | screenshot_before | screenshot_after |

amir20:
Looks good. I am going to test a little more and release this week.

homelab-alpha:
Take your time for testing.

Keep in mind that a container can utilize more than one CPU core. For example, a value of 180% indicates that one core is fully utilized at 100%, while a second core is running at 80%. In this case, the UI will display a `red status` because the CPU meter is capped at `max="100%"` percent. I am still investigating whether this maximum threshold can be dynamically adjusted to better represent multi-core usage.

amir20:
It probably can be. I have all that information. It will require some javascript.

homelab-alpha:
Let me know once you've made a decision about implementing this. If you need help with testing, feel free to reach out I’d be happy to assist. I’ve set up a few containers with multiple CPU cores, so I can run tests on my end if needed.

amir20:
I tested it. It seems to work. I'll release some time this week. 

> It probably can be. I have all that information. It will require some javascript.
> Let me know once you've made a decision about implementing this.

I was suggesting you can do it. 😁 I am pretty busy right now so I won't get to it.  I did something similar in 

https://github.com/amir20/dozzle/blob/7e2dd1c69ecf9b93ab4cfd21156ce415f66c4b16/assets/components/LogViewer/MultiContainerStat.vue#L58-L82

I think you can copy over some js and see if it works.  Perhaps the better solution would be to have CPU always between `0..100`.

homelab-alpha:
I’m going to tinker with the CPU meter 😁. 

But I also agree with you: the more complex the HTML5 <meter> implementation becomes, the harder it will be to maintain. 

I’ll keep you updated, and if it doesn’t work out, then that’s the end of it.

amir20:
Yup. It's also the dashboard that gets a lot of visibility.

PR #4035, opened 2025-07-13
fix: allows dots inside URLs (not at end)

Jeto143:
Dots are no longer allowed within URLs since https://github.com/amir20/dozzle/pull/4016.

This PR adjusts the regex to re-allow them inside URLs, but not at the end. See this page for examples:
https://regex101.com/r/GyEriu/4

Also took the opportunity to slightly clean up the regex by removing superfluous escapes.

Note: we could also optionally disallow `?` (question marks) at the end of URLs so that they're considered part of the sentence around it rather than the last char of the URL itself (even though it's technically valid).

Before:

image

After:

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.4
+          go-version: 1.24.5
           check-latest: true
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


amir20:
Thanks! Let me know when ready! I am pretty much used ChatGPT for the regex. So thanks for fixing it.

Jeto143:
> Thanks! Let me know when ready! I am pretty much used ChatGPT for the regex. So thanks for fixing it.

No worries! It's ready now, just had to bump the go version in the test workflow file to fix [this](https://github.com/amir20/dozzle/actions/runs/16251277366/job/45881292581).

Edit: It should probably be changed in the deploy workflow as well, I suppose.

amir20:
Merged. Going to also updated all deps. You can test with `master`. I'll release tomorrow.

PR #4038, opened 2025-07-15
fix: star conflicting with dropdown 

cavazzatommaso:
Resolved an issue where hovering near the star button would trigger the copy tooltip, obstructing the UI when logs are displayed beneath the header


amir20:
Thanks! merged

PR #4043, opened 2025-07-19
feat: group containers by `dev.dozzle.group` label

### LINKED ISSUE #3943: Group containers on left panel in compose mode

I have multiple compose nodes(without swarm mode) and there is more monitoring containers than other containers. I would like to move all monitoring containers into one group and show that group at the bottom of the left panel. This will make it easier to find "app" containers.

I imagine it like this

Image

Maybe use existing "custom container group" feature for this? Add "fork" icon(same as on the right side of the stack name) to show combined log output of custom container group. This will make possible to get rid of "two dots" at the very bottom of the left panel.

I tried to use [Custom Container Groups](https://dozzle.dev/guide/container-groups#custom-groups) (set `dev.dozzle.group: monitoring` in compose.yml) but it only combines log output into one feed

Image

Another alternative is to use different --project flag in compose. This will add extra logic to make sure I deploy container with proper --project flag.

ZenVoich:
fixes #3943

### DIFF
--- a/assets/models/Container.ts
+++ b/assets/models/Container.ts
@@ -80,7 +80,11 @@ export class Container {
   }
 
   get namespace() {
-    return this.labels["com.docker.stack.namespace"] || this.labels["com.docker.compose.project"];
+    return (
+      this.labels["dev.dozzle.group"] ||
+      this.labels["com.docker.stack.namespace"] ||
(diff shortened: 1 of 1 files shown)


amir20:
Thanks! Will release soon after I do a quick test.

PR #4091, opened 2025-08-29
feat: implement ability to disable avatars

### LINKED ISSUE #4090: FR: Gravatar toggle

Firstly,

Thank you for adding a forward proxy auth option. Since I already used authentik to proxy dozzle, it allowed me to easily display users with:
```yaml
environment:
      DOZZLE_AUTH_PROVIDER: forward-proxy
      DOZZLE_AUTH_HEADER_USER: X-Authentik-Username
      DOZZLE_AUTH_HEADER_EMAIL: X-Authentik-Email
      DOZZLE_AUTH_HEADER_NAME: X-Authentik-Name
```

However, I noticed that the avatar is loaded from Gravatar. I would rather not have an external connection to Gravatar for this purpose. Some sort of toggle would be great which would not request the avatar from Gravatar.

For example, perhaps if disabled via a toggle it would load a generic user icon instead or perhaps the name initial(s).

_No response_

uponminiature:
Resolves #4090.

This PR adds an option (argument or env var) to disable avatars entirely.
If the option is set to true, a user icon will be displayed in the place of the avatar instead and the avatar route will be disabled.

### DIFF
--- a/assets/components/Links.vue
+++ b/assets/components/Links.vue
@@ -14,7 +14,7 @@
 
     <dropdown class="dropdown-end" v-if="config.user">
       <template #trigger>
-        <template v-if="disableAvatars">
+        <template v-if="config.disableAvatars || !config.user.email">
           <material-symbols:person class="size-6" />
         </template>
         <template v-else>
@@ -29,12 +29,12 @@
           <div class="font-bold">
             {{ config.user.name }}
           </div>
-          <div class="text-sm font-light">
+          <div v-if="config.user.email" class="text-sm font-light">
             {{ config.user.email }}
           </div>
         </div>
-        <ul class="menu mt-4 p-0">
-          <li v-if="config.authProvider === 'simple'">
+        <ul v-if="config.authProvider === 'simple'" class="menu mt-4 p-0">
+          <li>
             <button @click.prevent="logout()" class="text-primary p-2">
               <material-symbols:logout />
               {{ $t("button.logout") }}
@@ -46,8 +46,6 @@
   </div>
 </template>
 <script lang="ts" setup>
-const { disableAvatars } = config;
-
 async function logout() {
   await fetch(withBase("/api/token"), {
     method: "DELETE",
--- a/assets/components/Links.vue
+++ b/assets/components/Links.vue
@@ -14,10 +14,15 @@
 
     <dropdown class="dropdown-end" v-if="config.user">
       <template #trigger>
-        <img
-          class="ring-base-content/60 size-6 max-w-none rounded-full p-px ring-1"
-          :src="withBase('/api/profile/avatar')"
-        />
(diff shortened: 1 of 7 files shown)


assets/components/Links.vue
```diff
@@ -14,10 +14,15 @@
 
     <dropdown class="dropdown-end" v-if="config.user">
       <template #trigger>
-        <img
-          class="ring-base-content/60 size-6 max-w-none rounded-full p-px ring-1"
-          :src="withBase('/api/profile/avatar')"
-        />
+        <template v-if="disableAvatars">
```
amir20, line 17: As I wrote in the original issue, I feel like this should just happen automatically if email is blank. Are you still providing an email address? Most people don't read docs so it would be good to have better defaults. Thoughts?

uponminiature:
@amir20 Can you review my latest commit? I have attempted to address it and tested it a bit.

Also, I noticed there was extra whitespace and I believe I have also fixed that by moving `v-if="config.authProvider === 'simple'"` to the ul parent and not only the logout button. This was unrelated to this PR, however.

amir20:
I think looks good. Thanks for doing it. And you were fast with these changes.

uponminiature:
> I think looks good. Thanks for doing it. And you were fast with these changes.

Alright, neat. Thank you as well for putting in the monumental amount of commits to make Dozzle in the first place. Always amazes me how much effort is put into open-source projects, and that those projects are free to use and modifiable - I am always grateful to the maintainers.

uponminiature:
> I think looks good. Thanks for doing it. And you were fast with these changes.

Also, would you be interested if I were to write up a PR which would add an opt-in "disable update checks" toggle? Unless this functionality already exists and I missed it. I think it would be useful for some, I am one to pin the docker images I pull to specific versions and check docker hub (or ghcr) weekly; speaking of, I should likely look into Diun to alert me of updates, haven't gotten around to that though.

amir20:
I have released it. 

> Also, would you be interested if I were to write up a PR which would add an opt-in "disable update checks" toggle? 

It doesn't exist. Before you send a PR I would want a feature request issue so that we can align on the expected implementation. I don't think it would make sense to just disable it and remove the drop down. I use that drop down to have other announcements.  

I see 2 options: 

1. Still check for latest version when clicking the drop down. This way the user explicitly taking an action. 
2. Disable all together and make it clear in the UI that it has been disabled.

uponminiature:
> > Also, would you be interested if I were to write up a PR which would add an opt-in "disable update checks" toggle?
> 
> It doesn't exist. Before you send a PR I would want a feature request issue so that we can align on the expected implementation. I don't think it would make sense to just disable it and remove the drop down. I use that drop down to have other announcements.

Alrighty, glad I asked before attempting an implementation. I'll fill out an issue soon™.

PR #4133, opened 2025-09-21
feat: Add user roles

dima-bzz:


amir20:
@dima-bzz could you provide an example `users.yml` so I can test it.

dima-bzz:
```yml
users:
  test:
    email: ""
    name: Test
    password: $2a$12$Bm4MgS5mrmgVExgE9Ba6q.OPdRrG0gssmu3ps8.X/hFc7XeOMyTWi #test
    roles: shell,actions,download
```

internal/auth/users.go
```diff
@@ -24,7 +25,9 @@ type User struct {
 	Name            string                    `json:"name" yaml:"name"`
 	Password        string                    `json:"-" yaml:"password"`
 	Filter          string                    `json:"-" yaml:"filter"`
+	Roles           string                    `json:"-" yaml:"roles"`
```
amir20, line 28: Maybe this is just a personal taste but I would call this `RolesConfigred` and the other one just `Roles`. It feels better to write `user.roles.Has(...)` than `user.userRoles.HasRole(...)`. 

Not a big deal. I could always change it on my side.

internal/web/index.go
```diff
@@ -45,12 +46,29 @@ func (h *handler) executeTemplate(w http.ResponseWriter, req *http.Request) {
 	user := auth.UserFromContext(req.Context())
 
 	if h.config.Authorization.Provider == NONE || user != nil {
+		enableShell := h.config.EnableShell
+		enableActions := h.config.EnableActions
+		enableDownload := true
+
+		if user != nil && user.UserRoles.Exists() {
```
amir20, line 53: I think this could be cleaned up a bit, eg

```
config["enableActions"] = h.config.EnableActions && user.UserRoles.HasRole(role.Shell) 
```

main.go
```diff
@@ -130,11 +130,11 @@ func main() {
 }
 
 func fileExists(filename string) bool {
-    _, err := os.Stat(filename)
-    if os.IsNotExist(err) {
-        return false
-    }
-    return err == nil
+	_, err := os.Stat(filename)
```
amir20, line 133: 🤨 what's going on here?

amir20 (changes requested):
I provided some quick feedback. Overall, it looks good! Thanks for doing it. I requested some changes. I can always go after and clean it up but it would delay launching it. 

Could you make these changes? And also update documentation at `docs/guide/authentication.md`.

internal/role/roles.go
```diff
@@ -0,0 +1,64 @@
...
+
+		if _, exists := seen[role]; !exists {
+			seen[role] = struct{}{}
+			roles = append(roles, role)
+		}
+	}
+	return roles
+}
+
+func (ur UserRoles) HasRole(role Role) bool {
```
amir20, line 57: If I am understanding this correctly, it is not possible to have no roles. Right? I wonder if none should be an option.

dima-bzz: If there are no roles, global parameters will be used

internal/role/roles.go
```diff
@@ -0,0 +1,64 @@
+package role
+
+import (
+	"strings"
+)
+
+type Role string
+
+const (
+	Shell    Role = "shell"
```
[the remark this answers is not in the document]
dima-bzz: I put it in auth at first, but I thought that auth refers to authorization. I can move it back to auth. Could you tell me how to properly convert a string to iota

amir20:
I think it's good now.

amir20 (approved):

amir20:
Try `master` and I'll release tomorrow.

amir20:
Ok i did testing. I think it's in a good shape now. I did notice a bug where roles won't be applied until _after_ user signs in again. Not an easy fix. 

I am going to release @dima-bzz. Thanks again.

PR #4135, opened 2025-09-22
fix: fixed container start time display after recreation

dima-bzz:
Before this fix, when Dozzle was opened the container start time was shown correctly.  
Снимок экрана 2025-09-22 в 08 55 10
However, if the container was recreated, the start time was set to `0001-01-01T00:00:00Z`,  
which in the interface appeared as **2026 years ago**.  
Снимок экрана 2025-09-22 в 08 57 58

The same incorrect value was also shown for stopped containers.

Снимок экрана 2025-09-22 в 14 21 26
Снимок экрана 2025-09-22 в 14 21 36

### DIFF
--- a/assets/components/ContainerPopup.vue
+++ b/assets/components/ContainerPopup.vue
@@ -5,13 +5,13 @@
         <th class="text-right font-light capitalize">STATE</th>
         <td class="font-semibold uppercase">{{ container.state }}</td>
       </tr>
-      <tr v-if="container.startedAt.getFullYear() > 0">
+      <tr v-if="container.startedAt.getFullYear() > 1">
         <th class="text-right font-light capitalize">STARTED</th>
         <td class="font-semibold">
           <RelativeTime :date="container.startedAt" />
         </td>
       </tr>
-      <tr v-if="container.state != 'running' && container.finishedAt.getFullYear() > 0">
+      <tr v-if="container.state != 'running' && container.finishedAt.getFullYear() > 1">
         <th class="text-right font-light capitalize">FINISHED</th>
         <td class="font-semibold">
           <RelativeTime :date="container.finishedAt" />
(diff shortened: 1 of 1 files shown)


amir20:
I don't think is the right implementation. Dozzle intentionally doesn't load the detail for stopped containers to improve performance.  You can see that here https://github.com/amir20/dozzle/blob/84ec399e50cae96a9bfd18e65ec8a6b7747a6755/internal/container/container_store.go#L86-L105

The code above only loads non-exited containers. Which means if you hover over a stopped container, you will see 2026 years ago. But if you click on it, I catch it and update only that container. That logic is here https://github.com/amir20/dozzle/blob/84ec399e50cae96a9bfd18e65ec8a6b7747a6755/internal/container/container_store.go#L175-L210

> However, if the container was recreated, the start time was set to 0001-01-01T00:00:00Z,
which in the interface appeared as 2026 years ago.

I'll have to test this because it shouldn't happen. What is the command you are running to recreate the container? Most likely the right event is not being captured at https://github.com/amir20/dozzle/blob/84ec399e50cae96a9bfd18e65ec8a6b7747a6755/internal/container/container_store.go#L260

This is all implemented for performance reason because if Dozzle had to load all containers (including stopped) it could spike CPU or wait a long time to load.

dima-bzz:
I just hid startedAt and finishedAt for stopped containers so that “2026 years ago” is not shown.
Снимок экрана 2025-09-22 в 22 00 05

When I open a container, the dates are displayed correctly.
Снимок экрана 2025-09-22 в 22 01 22

amir20:
> When I open a container, the dates are displayed correctly

So it works? 

I think the started dates is already hidden though. Going to close this.

dima-bzz:
I’m running the usual `docker compose up -d` command.

dima-bzz:
> > When I open a container, the dates are displayed correctly
> 
> So it works?
> 
> I think the started dates is already hidden though. Going to close this.

I don't hide them.

PR #4142, opened 2025-09-24
feat: add TLS support for web interface

dima-bzz:
Adds support for TLS/HTTPS in the web interface

### DIFF
--- a/internal/web/routes.go
+++ b/internal/web/routes.go
@@ -2,6 +2,7 @@ package web
 
 import (
 	"context"
+	"crypto/tls"
 	"io/fs"
 	"time"
 
@@ -10,6 +11,7 @@ import (
 
 	"github.com/amir20/dozzle/internal/auth"
 	"github.com/amir20/dozzle/internal/container"
+	"github.com/amir20/dozzle/internal/support/cli"
 	container_support "github.com/amir20/dozzle/internal/support/container"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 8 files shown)


amir20:
Hi @dima-bzz, 

Thanks for sending a PR. When submitting a PR like this, it's helpful to open an issue first to align on the scope and necessity. I have some feedback on the code, but before I provide that, I want to question this functionality.

1. I am unsure whether adding SSL support should be part of Dozzle. TLS/SSL is better suited for a proxy layer like Nginx or Traefik. Many have asked this question, and I consistently state that it is out of scope.  
2. Expanding Dozzle's functionality raises concerns because I frequently receive requests for help. If TLS is integrated into Dozzle, I will likely face more issues to troubleshoot, and I don't have the time for that.  
3. Additionally, I receive many pull requests that implement solutions tailored only to the author's use case. Dozzle supports Swarm, Kubernetes, and Podman, all of which need to be supported and documented.  

I am curious how you are using Dozzle. Are you using it for your own home network or being used in a company?

internal/web/routes.go
```diff
@@ -85,6 +87,14 @@ func CreateServer(hostService HostService, content fs.FS, config Config) *http.S
 		hostService: hostService,
 	}
 
+	certs, err := cli.ReadWebCertificates()
```
amir20, line 90: This is not good design. Calling cli package makes testing harder.  Instead `tls.Config` should be passed in `main.go`.

dima-bzz:
I use it both at home and in the company. I've been using dozzle for a long time, and I really like it.

amir20:
@dima-bzz I'm not sure what's happening, but communication has been challenging. You only responded to my last question [here](https://github.com/amir20/dozzle/pull/4142#issuecomment-3329134451). I will close this topic. Let's start a new discussion about the need for SSL. If enough people express interest or if adding this complexity to Dozzle proves valuable, we can reopen it. Based on the current discussion, not many users need this feature, and I prefer to keep Dozzle simple.

dima-bzz:
Good. For now, I’m just using it with Docker and regular Docker Compose. I’m currently setting up a central Dozzle with proxy auth and agents, and I’m planning to move to Kubernetes down the line.

dima-bzz:
I agree that SSL/TLS is usually better handled at the proxy level, for example via Nginx or Traefik, and that adding this functionality within Dozzle could increase your support burden. That’s why I appreciate you pointing out the need to consider Swarm, Kubernetes, and Podman.

Thanks again for your support!

amir20:
Great! Wasn't sure if you were in agreement.

PR #4147, opened 2025-09-25
docs: add SSO example for Pocket ID

mitchplze:
Working configuration for Pocket ID using oauth2-proxy container.

### DIFF
--- a/docs/guide/authentication.md
+++ b/docs/guide/authentication.md
@@ -372,3 +372,84 @@ services:
 ```
 
 After running the Dozzle container, configure the Application in Cloudflare Zero Trust dashboard by following the [guide](https://developers.cloudflare.com/cloudflare-one/applications/configure-apps/self-hosted-apps/).
+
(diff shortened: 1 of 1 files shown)


docs/guide/authentication.md
```diff
@@ -372,3 +372,84 @@ services:
 ```
 
 After running the Dozzle container, configure the Application in Cloudflare Zero Trust dashboard by following the [guide](https://developers.cloudflare.com/cloudflare-one/applications/configure-apps/self-hosted-apps/).
+
+### Setting up Dozzle with Pocket ID
+
+You must first setup a container to pass OpenID Connect authentication through your reverse proxy.
+
+Below is an example using [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy).
```
amir20, line 380: I wonder if there is a way to hide this by default? Auth page is getting huge and it would probably make sense to use `<summary>` here. What do you think?

PR #4150, opened 2025-09-26
fix: supports roles as json, comma or pipe sepearated values

amir20:


amir20:
@dima-bzz can you review? This is only the formatting.

dima-bzz:
I’m not entirely sure what the JSON and | are used for in this context. Could you clarify?

I think this was taken from here: [Dozzle Discussion #4118](https://github.com/amir20/dozzle/discussions/4118#discussioncomment-14464915)

amir20:
Right. So with this PR, `Remote-Roles: ["shell", "actions"]` is also supported. It is from that discussion [this comment](https://github.com/amir20/dozzle/discussions/4148#discussioncomment-14524241).

dima-bzz:
@amir20 I think on [roles.go#L37](https://github.com/amir20/dozzle/blob/master/internal/auth/roles.go#L37) it should be changed to Debug level; otherwise, it will create a lot of noise if there are other roles.

PR #4153, opened 2025-09-27
feat: support for dozzle_* roles

amir20:


dima-bzz:
I tested it with the proxy using dozzle_* groups, and everything works.
However, the documentation also needs to be updated: [authentication guide](https://github.com/amir20/dozzle/blob/master/docs/guide/authentication.md?plain=1#L164).

amir20:
Yep. Good call on docs. I'll come back to it before releasing.

Issue #4178, opened 2025-10-13
Agent segmentation fault after running for few days


Scarsz:
### 🔍 Check for existing issues

- [x] Completed

### How is Dozzle deployed?

Agents

### 📦 Dozzle version

v8.14.4

### ✅ Command used to run Dozzle

`docker run -d -v /var/run/docker.sock:/var/run/docker.sock -p 10.0.0.2:7007:7007 --name logs amir20/dozzle:latest agent`

### 🐛 Describe the bug / provide steps to reproduce it

Happens intermittently after container has been running for a few days

```
scarsz@redacted:~$ docker logs -f logs
{"level":"info","version":"v8.14.4","time":"2025-10-04T02:46:52Z","message":"Dozzle agent version v8.14.4"}
{"level":"info","version":"v8.14.4","time":"2025-10-04T02:46:52Z","message":"Agent listening on [::]:7007"}
{"level":"info","version":"v8.14.4","host":"redacted","time":"2025-10-04T09:05:03Z","message":"stopped container stats collector"}
{"level":"info","version":"v8.14.4","host":"redacted","time":"2025-10-05T11:31:10Z","message":"stopped container stats collector"}
{"level":"info","version":"v8.14.4","host":"redacted","time":"2025-10-05T20:14:39Z","message":"stopped container stats collector"}
{"level":"info","version":"v8.14.4","host":"redacted","time":"2025-10-07T02:07:55Z","message":"stopped container stats collector"}
{"level":"info","version":"v8.14.4","host":"redacted","time":"2025-10-07T15:23:00Z","message":"stopped container stats collector"}
{"level":"warn","version":"v8.14.4","userLabels":{"label":["redacted"]},"time":"2025-10-07T23:15:30Z","message":"no containers found with user labels"}
{"level":"warn","version":"v8.14.4","userLabels":{"label":["redacted"]},"time":"2025-10-07T23:15:32Z","message":"no containers found with user labels"}
{"level":"warn","version":"v8.14.4","userLabels":{"label":["redacted"]},"time":"2025-10-07T23:15:32Z","message":"no containers found with user labels"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xaccc77]

goroutine 5034 [running]:
github.com/amir20/dozzle/internal/agent.logEventToPb(0x0)
	/dozzle/internal/agent/server.go:356 +0x17
[...]

PR #4183, opened 2025-10-14
fix: fixes a nil pointer that was happening in the agent

amir20:
Fixes https://github.com/amir20/dozzle/issues/4178


amir20:
@claude is there  a way to make these comments shorter going forward?

amir20:
@claude Ultra-brief and how do I configure so all PRs going forward are brief?

PR #4193, opened 2025-10-17
chore: pins node to v24

amir20:


amir20:
@claude update to latest 24.x patch

PR #4210, opened 2025-10-24
fix: scroll-jump by making dropdown position dynamic

### LINKED ISSUE #4204: When hovering over the menu of the last log in the list, the display is jumpy

- [x] Completed

### How is Dozzle deployed?

Agents

### 📦 Dozzle version

v8.14.5

compose.yml

```yaml
services:
  dozzle:
    image: amir20/dozzle:latest
    hostname: dozzle
    container_name: dozzle
    volumes:
    environment:
      DOZZLE_REMOTE_AGENT: 1.2.3.4:7007
```

I noticed it while searching but it looks like it does something similar in the regular logs view, but when hovering over the last log item and attempting to hover over the popup menu icon that shows on the left side the display becomes jumpy because it scrolls out of the way which then hides the menu which then places the icon back under the cursor and then repeats.  This seems to be more obvious when the last log item is not at the very bottom of the screen.

### 💻 Environment

This is more a browser/html issue than server. But let me know if you really need them.

I don't think server logs will be needed for this.

Image

akash-ramaswamy:
Fixes #4204


amir20:
I'll check this out today. I want to ensure there is no performance loss, which there shouldn't be since it's only on mouseover.

PR #4216, opened 2025-10-25
fix: make scrollable classes conditional in ScrollableView

akash-ramaswamy:
The 'snap-y overflow-auto' classes are now applied to the `<main>` element only when 'scrollable' is true. issue #4204

### DIFF
--- a/assets/components/ScrollableView.vue
+++ b/assets/components/ScrollableView.vue
@@ -6,7 +6,7 @@
     >
       <slot name="header"></slot>
     </header>
-    <main :data-scrolling="scrollable ? true : undefined" class="snap-y overflow-auto">
+    <main :data-scrolling="scrollable ? true : undefined" :class="{ 'snap-y overflow-auto': scrollable }">
       <div class="invisible relative md:visible" v-show="scrollContext.paused">
         <div class="absolute top-4 right-44">
           <ScrollProgress


akash-ramaswamy:
I hope this completely fixes the scroll-jump issue. I have tested this to the best of my knowledge. It works perfectly fine on mobile too.

But @amir20 wrt below context, Was there any purpose you didn't make the `<main>` scroll classes conditional?
https://github.com/amir20/dozzle/blob/313cde3ad11697c1ce834ae2688def2399c43efb/assets/components/ScrollableView.vue#L9

amir20:
I don't think this is right because the snappy class is needed all the time. Otherwise, the auto scroll won't work. I'll do some testing.

amir20:
I am not sure what this fixes. It seems to already be working on `master`. 


https://github.com/user-attachments/assets/dfd065b9-b8bd-4e68-a874-aa09e20b519d

akash-ramaswamy:
yeah, this is fixed. But if there are like only 3-5 logs, that jump issue occurs.

amir20:
Can you send a video? I am not sure I understand.

akash-ramaswamy:
https://github.com/user-attachments/assets/81cc3122-ec1e-4500-84ac-1528bacbb220

amir20:
Oh I see. Seems like this PR might break other things then. 

Maybe a better fix would be if content is larger than view port? I can check later.

amir20:
Confirmed that this is also broken with multiple vertical panes. I think it needs a better solution.

amir20:
See https://github.com/amir20/dozzle/pull/4217

akash-ramaswamy:
This seems to be a simple and elegant solution. Never thought this way! More to learn from this repo @amir20. Dozzle never failed to amuse me.

PR #4261, opened 2025-11-25
feat: add context navigation controls for search results

Amrrx:
When viewing a log line via "see in context" from search results, navigation buttons will appear on the highlighted target line allowing users to scroll ±10 or ±50 lines from the current position.

Summary

  - Adds navigation buttons (±10 and ±50 lines) to highlighted log lines when viewing search results in context
  - Allows users to quickly scroll through logs relative to their current position
  - Navigation controls only appear on lines highlighted via "see in context" from search

### DIFF
--- a/assets/components/LogViewer/LogItem.vue
+++ b/assets/components/LogViewer/LogItem.vue
@@ -16,7 +16,7 @@
         v-if="showTimestamp"
         :date="logEntry.date"
         class="shrink-0 select-none"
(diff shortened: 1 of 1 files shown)


amir20:
The implementation looks correct. I tested it, and it works. However, using `Array.from` may have performance implications, as it can be costly in the browser. I am questioning the use case. What problem does this PR solve? What is the difference between scrolling up with the mouse and clicking the up button? 

Otherwise, looks good.

Amrrx:
You were right

PR #4280, opened 2025-12-09
fix: do not hide ContainerActionsToolbar component on mobile

adrienlegall:
This PR makes the ContainerActionsToolbar component visible on mobile devices by removing the `max-md:hidden` class.

### DIFF
--- a/assets/components/ContainerViewer/ContainerLog.vue
+++ b/assets/components/ContainerViewer/ContainerLog.vue
@@ -9,7 +9,7 @@
           v-if="container.state === 'running'"
         />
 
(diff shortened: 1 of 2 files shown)


amir20:
I wouldn't say this is a bug. This is intentional. In the past, search didn't work really well. The textbox would freeze. Not sure if that's still an issue. Have you tested it?

amir20:
Was able to test it out. 
Screenshot 2025-12-10 at 7 34 05 PM

Search is a little broken as it makes the page jump up and down. And for longer names, they get cut off. 

To merge this, I think it needs a little more. I won't have time right now. 

Let me know if you can fix it.

Issue #4350, opened 2026-01-16
Option to use compose service name instead of container name


aasmoe:
### Describe the feature you would like to see

Dozzle beautifully groups containers by stack, but since my setup requires not setting explicit container names, all containers are listed with project name prefixed. This is not so beautiful.

<img width="336" height="905" alt="Image" src="https://github.com/user-attachments/assets/471adeb7-36a7-4387-aeec-208a01f11ad1" />

Could there be an option to use the docker compose service names as display name instead of container name?

<img width="249" height="914" alt="Image" src="https://github.com/user-attachments/assets/6c1a0736-b866-4038-a416-7482dd08224c" />

### Describe how you would like to see this feature implemented

`docker comopse -p <stackname> ps` returns service name along with container name. There are probably other ways to get it with the API. 

In the UI there could be toggle for using this type of container display name.

### Describe any alternatives you've considered

- Use `dev.dozzle.name` labels. This solves the problem! But requires manually setting labels for each container. 
- Another approach to removing stack name could be to detect if the stack name is prefixed and remove it before listing the container.

amir20:
That's an interesting idea. I wonder how `docker compose -p ... ps` works. I don't see a service name in the API. I could strip the stack name myself, but I hope there's a better way.

amir20:
Seem like `com.docker.compose.service` label is the right thing to use 🤔

aasmoe:
Cool, I did a small test with that:
```go
func newContainer(c docker.Summary, host string) container.Container {
	name := c.Labels["com.docker.compose.service"]
```

The result is exactly like the second picture above, i.e. perfect. I guess that will not be a value in all running containers, only those ran using docker compose. So it would need fallback.

amir20:
That's right. But I think the logic needs to be in the UI and not backend. If a container is grouped under the stack name, then it should use service name. Which is a bit different. I'll keep this open until I can get to it. If you time a PR would be appreciated too.

amir20:
Can you test #4358 with `amir20/dozzle:pr-4358`? I realized I already had some logic for this and with this change, I might have broken something else. So a through testing would be great. 

LMK.

amir20:
Moving to discussion. People can upvote or suggest better options. Closed the PR as it turns out `container_name` needs to be respected and there is no easy way.

PR #4358, opened 2026-01-18
feat: improves stacknames and containers by using service name inside stack

amir20:
Fixes https://github.com/amir20/dozzle/issues/4350

### DIFF
--- a/assets/components/HostMenu.vue
+++ b/assets/components/HostMenu.vue
@@ -60,7 +60,7 @@
     </template>
     <template #right>
       <ul class="containers menu w-full p-0 [&_li.menu-title]:px-0">
-        <li v-for="{ label, containers, icon, isNamespaceGroup } in menuItems" :key="label">
+        <li v-for="{ label, containers, icon } in menuItems" :key="label">
           <details :open="!collapsedGroups.has(label)" @toggle="updateCollapsedGroups($event, label)">
             <summary class="text-base-content/80 font-light">
               <component :is="icon" />
@@ -93,7 +93,7 @@
                       :data-state="item.state"
                     ></div>
                     <div class="truncate">
-                      {{ isNamespaceGroup && item.serviceName ? item.serviceName : item.name }}
+                      {{ item.name }}
                     </div>
                     <ContainerHealth :health="item.health" />
                     <span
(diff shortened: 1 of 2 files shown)


aasmoe:
I have done some testing. 

Containers that have `container_name` configured are still displayed using service name. There could be a priority order dozzle label -> container_name -> service name -> name. 

But looking at the container json using docker inspect I cant find a reliable way to tell if a container has a set `container_name` other than matching `com.docker.compose.replace` to the pattern `com.docker.compose.service + com.docker.compose.container-number` (like dozzle-1). If it follows that pattern, `container_name` has probably not been set.

Other than this, Dozzle seems to work as expected (I dont use swarm).

amir20:
Oh hmm. 

> Containers that have container_name configured are still displayed using service name. There could be a priority order dozzle label -> container_name -> service name -> name.

I am not sure if it's possible to do both. I can either use `container.name` which also works with `container_name` or `service.name`. But Docker API doesn't return anything indicating it has been overridden. 

> But looking at the container json using docker inspect I cant find a reliable way to tell if a container has a set container_name other than matching com.docker.compose.replace to the pattern com.docker.compose.service + com.docker.compose.container-number (like dozzle-1). If it follows that pattern, container_name has probably not been set.

This sounds complicated. Because with swarm it's a little different. 

I am inclined to keep the behavior as is. Seems like it's more important to respect `container_name`. 

What do you think?

aasmoe:
> I am inclined to keep the behavior as is. Seems like it's more important to respect `container_name`.

I agree its more important to respect `container_name` and pattern-matching is too complicated.

To avoid messing with `container_name`-configured containers, there could be an option to use service names only if user prefers to.

image


Then only use service names when that setting is enabled, otherwise use the name value as before.


```ts
    if (this.isSwarm) {
      return this.labels["com.docker.swarm.service.name"];
    }
    if (serviceNames.value) {
      return this.labels["com.docker.compose.service"] || this._name;
    }
    return this._name;
  }
```

I like this solution, even if its not as sleek as adding service names to a prioritization order. Importantly, it would not change the way users' Dozzle looks like unless they enable the option.

If the feature is worth the added setting, I can submit a PR, if you want.

amir20:
To be honest, I try to keep the settings to an absolute minimum. It adds too much testing and cognitive load for users to have too many options. Most of the feature requests I get are already in the setting. I'd say let's revisit this if others also have the same request.

PR #4379, opened 2026-01-23
fix: Avoid fatal on invalid container filters in auth

Mirza-Samad-Ahmed-Baig:
Prevents Dozzle from terminating when invalid container filters are supplied via forward-proxy headers or JWT claims.

Returns 400 Bad Request for invalid forward-proxy filters and treats invalid JWT filters as unauthenticated.


amir20:
Makes sense. I am working on v10 right now though. So I'll merge this in master and figure out how to release later.

PR #4382, opened 2026-01-24
fix: use io.Copy for exec TTY output instead of stdcopy

robertbreker:
Without TTY, Docker multiplexes stdout/stderr by prefixing each chunk with an 8-byte header (stream type + length). stdcopy.StdCopy parses these headers.

With TTY enabled, output is raw terminal data with no headers. stdcopy.StdCopy misinterprets the first bytes as a header, reads a bogus length, and blocks forever waiting for bytes that never arrive.

Exec always uses TTY mode, so io.Copy is correct (as Attach already does when c.Tty is true).

### DIFF
--- a/internal/support/docker/docker_service.go
+++ b/internal/support/docker/docker_service.go
@@ -208,7 +208,8 @@ func (d *DockerClientService) Exec(ctx context.Context, c container.Container, c
 	})
 
 	wg.Go(func() {
-		if _, err := stdcopy.StdCopy(stdout, stdout, session.Reader); err != nil {
+		// TTY mode outputs raw bytes without Docker's multiplexing headers.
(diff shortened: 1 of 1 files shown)


amir20:
Have you tested master? It has changed a lot since `v9.0.2`? I am currently in the process of adding a lot of features. 

I have never noticed exec not working.

rbreker-godaddy:
Hi Amir,

Yes - this problem reproduced on master for me.

I initially wondered if this was Podman-specific -
However, Attach() already correctly checks c.Tty and uses io.Copy for TTY mode. Exec() was just missing the same handling.

amir20:
Ok as long as you tested in `master` then I am 👍. I have made so many changes in master that I was worried it might be outdated.  I am currently working on a big feature for notifications. Will merge this. But won't release for a couple weeks.

PR #4383, opened 2026-01-24
fix: detect agent stream errors and reconnect

robertbreker:
Agents did not reconnect after restart, regardless of method used (podman restart, or podman stop/start). Streams broke but the connection appeared healthy, so no reconnection was triggered.

Adds Healthy() to ClientService interface. Agent sets healthy=false when any stream (stats, events, containers) fails. RetryAndList() checks health and reconnects unhealthy agents.

### DIFF
--- a/internal/agent/client_test.go
+++ b/internal/agent/client_test.go
@@ -89,6 +89,10 @@ func (m *MockedClientService) Exec(ctx context.Context, c container.Container, c
 	return args.Error(0)
 }
 
+func (m *MockedClientService) Healthy() bool {
+	return true
+}
+
 var wantedContainer = container.Container{}
 
 func init() {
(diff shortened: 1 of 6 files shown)


amir20:
Doesn't gRPC automatically reconnect? It's not instant but it does eventually.

rbreker-godaddy:
Based on my testing, gRPC may reconnect the underlying HTTP/2 connection, but not the streaming RPCs themselves.
When the agent blips, stream.Recv() returns an error and the stream is done. The StreamStats, StreamEvents goroutines exit with an error.

amir20:
Ok I reviewed this one. Something doesn't make sense. The stats stream is for per browser session. when `request.Context()` ends, it actually gets disconnected. 

I think something else is going on here. 

I think it needs a little more testing.

robertbreker:
Going to close this one out - traced it back to a different root cause.
Rootless Podman wasn't providing a stable host ID, and no amount of engine-id configuration would fix it.
Works fine on rootless Docker.
Amir - sorry for the false alarm, and appreciate you looking at it!

Issue #4390, opened 2026-01-27
change hardcoded downloaded zip log file name


FrancYescO:
### Describe the feature you would like to see

change
`filename=container-logs-%s.zip`
to
`filename=<containername>-logs-%s.zip`

and maybe 
`filename=<stackname>-logs-%s.zip`

if downloading the aggregate

amir20:
I like it! Not sure I'll get to it any time soon.

aatif101:
Hi, I'd like to work on this feature!

Currently:
The download filename is hardcoded as `container-logs-<timestamp>.zip` regardless of what's being downloaded [[download.go: line:114](https://github.com/amir20/dozzle/blob/master/internal/web/download.go#L114)]

Proposed implementation:

**Single container download:** Use `<container-name>-logs-<timestamp>.zip`

The container name is already accessible through `containerService.Container.Name`, so this is basically just swapping out the hardcoded prefix with the actual container name.

**For multi-container downloads where all containers belong to the same stack/project, we can generate:** `<stack-name>-logs-<timestamp>.zip`

For stack detection, we can extract labels from each container and resolve the stack name using the same priority order used on the frontend, so behavior stays consistent across UI and backend:

1. `dev.dozzle.group`
2. `coolify.projectName`
3. `com.docker.stack.namespace`
4. `com.docker.compose.project`

The idea here is:

* Collect candidate stack names from all selected containers
* If they all resolve to the same value -> use that as the filename prefix
* If they differ or don't exist -> treat it as a mixed selection

**For mixed containers (no shared stack / inconsistent labels), we fall back to:** `merged-logs-<timestamp>.zip`

**Sanitize filenames to replace any invalid characters (like / or :) with underscores**

The container info is already collected before setting the header, so this is mostly about determining the appropriate name based on the number of containers and their shared labels.

Let me know if this approach sounds good. I can implement this unless there's an objection.

amir20:
1. Container name makes sense
2. It's hard to come up with a good name for multiple containers. The problem is that it can be based on many filters and each one having their own set of containers. So two ideas: 
a. The front end can suggest a name because a the time of downloading the UI knows the state 
b. Just say `<x>-containers-logs-<timestampe>.zip`

aatif101:
That's a good point, I hadn't thought about all the different filter contexts that could trigger a download.

[...]

PR #4422, opened 2026-02-09
feat(download): generate dynamic filenames for log downloads

aatif101:
Addresses #4390

### What changed
Dozzle currently downloads log archives with a generic `container-logs-<timestamp>.zip` filename.

This PR generates context-aware filenames:
- **Single container**: `<container-name>-logs-<timestamp>.zip`
- **Multiple containers**: `<count>-containers-logs-<timestamp>.zip`

### Example
- `test-container-logs-2026-02-09T15-05-50`

### Notes
The UI can have many filter contexts (stack/service/merged view). This PR keeps backend naming simple for multi-container downloads by using container count.

### Testing
- Download logs for a single container  -> filename uses container name
- Download logs for multiple containers -> filename uses container count

### DIFF
--- a/internal/web/download.go
+++ b/internal/web/download.go
@@ -17,6 +17,15 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+// sanitizeFilename replaces characters that are invalid in filenames
+func sanitizeFilename(name string) string {
+	replacer := strings.NewReplacer(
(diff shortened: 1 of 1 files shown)


amir20:
This looks fine. But I think it can be better to make it nicer. So I'll review it again and see if I can improve it.

amir20:
See https://github.com/amir20/dozzle/pull/4428

PR #4429, opened 2026-02-11
fix!: Removes the sha256 support. Dozzle now does a fatal exit

amir20:


jonocairns:
@amir20 - how should we navigate this change if we have previously set up a user with the old hash?

amir20:
Just regenerate using the generate command.

PR #4452, opened 2026-02-16
feat: add metric-based alerts for container CPU/memory thresholds

dhaval-patel262:
## Summary

- Adds the ability to trigger webhook notifications when container resource thresholds are exceeded (e.g., CPU > 80%, memory > 90%)
- Extends the existing notification system by bridging the stats stream (already used for UI display) with the alert/dispatch pipeline
- Includes per-container cooldown to prevent notification flooding

## Motivation

Dozzle already streams real-time CPU, memory, and network metrics to the UI, but the notification system only supports log-pattern matching. This PR adds metric-based alerting so users can get notified (e.g., via Telegram webhook) when containers exceed resource thresholds.

## Changes

**Backend (Go):**
- `types/notification.go` — New `NotificationStat` struct, `Stat` field on `Notification`, `MetricExpression`/`Cooldown` on `SubscriptionConfig`
- `internal/notification/types.go` — `MetricExpression`, `Cooldown`, `MetricProgram` on `Subscription` + `MatchesMetric()`, cooldown methods
- `internal/notification/stats_listener.go` — **New file** - `ContainerStatsListener` subscribes to stats from all clients
- `internal/notification/manager.go` — `processStatEvents()` goroutine with per-container cooldown logic
- `internal/support/docker/multi_host_service.go` — Wires `ContainerStatsListener`, propagates new fields
- `internal/support/cli/agent_command.go` — Updated `NewManager` call
- `internal/web/notifications.go` — API types and handlers support `metricExpression` + `cooldown`

**Frontend (Vue/TypeScript):**
- `assets/types/notifications.ts` — Added `metricExpression`, `cooldown`, `metricError` fields
- `assets/composable/exprEditor.ts` — Added `createMetricHints()` with cpu/memory completions
- `assets/components/Notification/AlertForm.vue` — Log/Metric alert type toggle, metric expression editor, cooldown input
- `assets/components/Notification/AlertCard.vue` — Displays metric expression and cooldown for metric alerts

## How It Works

1. `ContainerStatsListener` subscribes to the existing `DockerStatsCollector` stats stream
2. `Manager.processStatEvents()` evaluates each stat against metric subscriptions
3. If `MetricExpression` matches (e.g., `cpu > 80`) and container matches `ContainerExpression`, and cooldown has elapsed → dispatches notification via existing `WebhookDispatcher`
4. Per-container cooldown tracking prevents flooding (default 5 minutes)

## Test plan

- [ ] Create a webhook destination pointing to Telegram bot API
- [ ] Create a metric alert with expression `cpu > 5` (low threshold for testing)
- [ ] Verify Telegram message arrives when a container exceeds threshold
- [ ] Verify cooldown prevents repeat notifications within the cooldown period
- [ ] Verify existing log-based alerts still work unchanged
- [ ] Run `go test ./...`

### DIFF
--- a/internal/notification/manager.go
+++ b/internal/notification/manager.go
@@ -430,7 +430,7 @@ func (m *Manager) processLogEvent(logEvent *container.LogEvent) {
 		notification := types.Notification{
 			ID:        fmt.Sprintf("%s-%d", c.ID, time.Now().UnixNano()),
 			Container: notificationContainer,
-			Log:       notificationLog,
+			Log:       &notificationLog,
 			Subscription: types.SubscriptionConfig{
 				ID:                  sub.ID,
 				Name:                sub.Name,
@@ -456,22 +456,21 @@ func (m *Manager) processStatEvents() {
 		select {
 		case <-m.ctx.Done():
 			return
-		case stat := <-m.statsListener.StatChannel():
-			m.processStatEvent(stat)
+		case event := <-m.statsListener.Channel():
+			m.processStatEvent(event)
 		}
 	}
 }
 
 // processStatEvent processes a single stat event and sends notifications for matching metric subscriptions
-func (m *Manager) processStatEvent(stat container.ContainerStat) {
+func (m *Manager) processStatEvent(event ContainerStatEvent) {
 	notificationStat := types.NotificationStat{
-		CPUPercent:    stat.CPUPercent,
-		MemoryPercent: stat.MemoryPercent,
-		MemoryUsage:   stat.MemoryUsage,
+		CPUPercent:    event.Stat.CPUPercent,
+		MemoryPercent: event.Stat.MemoryPercent,
+		MemoryUsage:   event.Stat.MemoryUsage,
 	}
 
-	// Lazily resolved once per stat event, shared across all subscriptions
-	var notificationContainer *types.NotificationContainer
+	notificationContainer := FromContainerModel(event.Container, event.Host)
 
 	m.subscriptions.Range(func(_ int, sub *Subscription) bool {
 		// Skip disabled or non-metric subscriptions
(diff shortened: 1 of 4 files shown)


amir20 (commented):
## PR #4452: Metric-based alerts for container CPU/memory thresholds

### Critical Issues

- **Race condition on `MetricCooldowns`**: `Subscription.MetricCooldowns` is initialized lazily in `SetMetricCooldown()` (`types.go`). If two goroutines call `SetMetricCooldown` concurrently on a new subscription, both could see `nil` and race on assignment. Initialize `MetricCooldowns` in `AddSubscription`/`loadSubscription`/`ReplaceSubscription` instead.

- **`FindContainerWithHost` called per stat per subscription** (`manager.go:~500`): `processStatEvent` does a linear scan of all clients for every matching subscription on every stat tick. Stats fire frequently — this will be expensive. Consider caching container→host mapping or looking up the container once outside the subscription loop.

- **Unbounded goroutine per notification** (`manager.go:~549`): `go m.sendNotification(d, notification, sub.DispatcherID)` inside `Range` — under high stat volume with many matching subscriptions, this spawns unbounded goroutines. The `sendSem` semaphore helps, but goroutine creation itself isn't bounded. Consider using a worker pool or at least a goroutine limiter.

- **Dummy event channel leak** (`stats_listener.go:31-32`): `dummyEvents` channel with buffer 100 is created per client but events are only drained while `ctx` is alive. If events arrive faster than draining, the producer could block. More importantly, the goroutine draining events captures `dummyEvents` by closure but the channel is never closed — goroutine lives until ctx cancellation, which is fine, but the pattern is fragile.

### Bugs

- **`canSave` allows empty `logExpression` for log alerts** (`AlertForm.vue:~179`): For log type, validation is `base && !logError.value` — but `logExpression` can be empty and `logError` will be `null`, so a log alert with no log expression passes validation. The original code had the same issue, but this is a good time to fix it.

- **Cooldown `int` type mismatch** (`manager.go:~253`): `UpdateSubscription` casts cooldown with `value.(int)`, but JSON unmarshaling typically produces `float64` for numbers. This type assertion will silently fail, ignoring cooldown updates via PATCH.

### Suggestions

- **Duplicated expression compilation**: The `expr.Compile(sub.MetricExpression, expr.Env(...))` block is repeated 4 times (`AddSubscription`, `ReplaceSubscription`, `UpdateSubscription`, `loadSubscription`). Extract a `compileMetricExpression` helper.

- **Hardcoded strings**: `AlertCard.vue` has untranslated strings ("Metric", "Cooldown") and `AlertForm.vue` has ("Alert Type", "Log Alert", "Metric Alert", "Metric Expression", etc.). These should use `$t()` i18n keys like the rest of the notification UI.

- **`dozzle` image skip is brittle** (`manager.go:~509`): `strings.Contains(c.Image, "dozzle")` will match any image with "dozzle" in the name. This should match the same pattern used elsewhere in the codebase for self-filtering.

- **No tests**: 477 additions with zero test coverage. At minimum, unit tests for `MatchesMetric()`, `IsMetricCooldownActive()`, and `processStatEvent` logic.

amir20:
@dhaval-patel262 I did a review Claude 🥳. I am mostly concerned about performance problems because if not fixed then it can take down people's host. 

> Dummy event channel leak (stats_listener.go:31-32): dummyEvents channel with buffer 100 is created per client but events are only drained while ctx is alive. If events arrive faster than draining, the producer could block. More importantly, the goroutine draining events captures dummyEvents by closure but the channel is never closed — goroutine lives until ctx cancellation, which is fine, but the pattern is fragile.

That seems weird to need that. 

I'll have to play around with this PR to make sure it works. It will take me a while.

amir20:
There is a lot going on with this PR that doesn't seem right.  I am cleaning it up at https://github.com/amir20/dozzle/pull/4454. So I'll close this one.

PR #4466, opened 2026-02-21
fix: use coolify.serviceName for unique container names in multi-container apps

### LINKED ISSUE #4432: The new default Coolify labels appear less optimal for applications with multiple containers

Hello 👋 

In v10, a nice little QoL change was added which uses the `coolify.resourceName` if present as a label on the container to name it in the UI. This works well for applications with just one running container e.g. Dockerfile applications. It is however less than optimal for applications with multiple containers, e.g. a docker compose application. Each container in an application with multiple containers have the label `coolify.resourceName`, which will be identical to the name of the whole application, so they all show up with the same name in the UI.

An example of how it currently looks in my UI.

Image

Example of coolify labels that can be used for naming, on an application with multiple containers

```json
"coolify.name": "web-dg8oo4gwkssc8ks80wogs44k",
"coolify.resourceName": "monitor-glitchtip",
"coolify.service.subName": "glitchtip-website",
"coolify.service.subType": "application",
"coolify.serviceName": "glitchtip-website",
"coolify.type": "service",
```

I propose that instead of always defaulting to the `coolify.resourceName`, it should use the `coolify.service.subName` instead when present, which is only present in these kind of application types (AFAIK) which are then named individually.

I think `coolify.serviceName` would work as well, it appears to function as `coolify.service.subName` if present, otherwise it is `coolify.resourceName`.

javierperezm:
## Summary

- Use `coolify.serviceName` instead of `coolify.resourceName` as fallback for container display name
- `resourceName` is identical for all containers in a multi-container Coolify app, making them indistinguishable
- `serviceName` is unique per sub-service, falling back to `resourceName` for single-container apps

Fixes #4432

## Test plan

- [x] Existing label priority tests updated and passing
- [x] Verify with multi-container Coolify app that containers show distinct names

### DIFF
--- a/docs/guide/container-names.md
+++ b/docs/guide/container-names.md
@@ -32,7 +32,7 @@ services:
 
 If you're using [Coolify](https://coolify.io/), Dozzle automatically recognizes Coolify's labels as fallbacks:
 
-- `coolify.resourceName` → Used as container name if `dev.dozzle.name` is not set
+- `coolify.serviceName` → Used as container name if `dev.dozzle.name` is not set
 - `coolify.projectName` → Used for grouping if `dev.dozzle.group` is not set
 
 No additional configuration is needed for Coolify deployments.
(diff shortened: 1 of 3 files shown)


amir20:
Thanks!

PR #4601, opened 2026-04-08
Bundle docker-compose binary in scratch image

amir20:
Bundle `docker/compose-bin` v2.36.0 into the final scratch image so that `docker-compose` is available alongside the `dozzle` binary.

- Uses `--platform=$TARGETPLATFORM` on the compose stage to ensure the correct architecture is pulled during multi-arch builds.
- Copies `docker-compose` directly from the compose stage into `/usr/local/bin/docker-compose` in the final scratch image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ COPY public ./public
 RUN pnpm build
 
 # Grab docker-compose binary
-FROM docker/compose-bin:v2.36.0 AS compose
+FROM --platform=$TARGETPLATFORM docker/compose-bin:v2.36.0 AS compose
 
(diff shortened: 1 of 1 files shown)


amir20:
Addressed:
- Platform mismatch: compose stage now pinned with `--platform=$TARGETPLATFORM` (4b5576c).
- PR description: updated, UPX reference removed.
- UPX: dropped entirely (f2bbf89).

Keeping docker-compose bundled intentionally — it's being shipped for use by dozzle cloud tooling / container action workflows even though no `exec.Command` call exists yet in the tree. The size trade-off is accepted.

---

PR #4607, opened 2026-04-10
fix: Fix memory stats not showing on cgroup v2

utafrali:
Memory stats weren't showing on cgroup v2 because the Usage field could be 0. Added a fallback to use the anon stat when that happens, and removed the guard that was preventing stats from being sent.


amir20:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #4608, opened 2026-04-10
feat: add PWA support with dynamic base URL

amir20:
## Summary
- Serve `manifest.webmanifest` dynamically from Go backend, injecting the configured `DOZZLE_BASE` into `start_url`, `scope`, and icon paths
- Serve a minimal `sw.js` from `{base}/sw.js` — its scope is automatically `{base}/`, solving the dynamic base URL problem
- Register the service worker in the frontend via `withBase("/sw.js")`
- The SW caches Vite's hashed static assets (cache-first) and uses network-first for everything else

## Test plan
- [ ] `make dev` — verify manifest loads at `/manifest.webmanifest` with correct paths
- [ ] DevTools > Application > Service Workers — verify SW registers
- [ ] Test with `DOZZLE_BASE=/test` — verify manifest scope and SW scope update correctly
- [ ] Check PWA install prompt appears in browser
- [ ] `make test` and `pnpm test` — existing tests pass


internal/web/sw.go
```diff
@@ -0,0 +1,51 @@
+package web
+
+import "net/http"
+
+const serviceWorkerJS = `
+const CACHE_NAME = "dozzle-v1";
```
amir20: Should this be dozzle.version here instead?

PR #4649, opened 2026-04-23
feat: support grouped remote agents

nomah4:
## Summary

- Adds host grouping UX: hosts can be assigned a group via `host:port|name|group` endpoint format, displayed as collapsible sections in the sidebar
- Adds collapse/expand toggle per host group and a global collapse-all action
- Adds host-group merged log view (`/host-group/[name]`)
- Fixes offline agents not appearing in their group: `failedAgents` now parses the pipe-delimited endpoint to extract `name` and `group` fields instead of using the raw endpoint string

## Bug fix detail

When an agent fails to connect during startup, it was stored in `failedAgents []string` with only the raw endpoint. In `Hosts()`, the offline placeholder host was built without parsing `host:port|name|group`, so `Name` was the full endpoint string and `Group` was empty — causing the host to render outside its group.

Fix: parse the endpoint in `RetriableClientManager.Hosts()` the same way `agent.parseEndpoint` does.

## Test plan

- [ ] Configure a remote agent with group: `--remote-agent 1.2.3.4:7007|my-node|production`
- [ ] Take the agent offline
- [ ] Verify the offline host still appears under the `production` group in the sidebar with its name `my-node` (not the raw endpoint string)
- [ ] Verify collapse/expand of host groups persists across page reloads


amir20:
Let me know when this is ready to review. A few notes: 

- Make sure to also update `docs` with proper instructions to use this 
- Provide screenshots 
- Make sure everything builds with no typescript errors

One minor notes, it's already possible to show containers groups at host level. So I am not sure what this is solving except to do multiple groups together.

amir20:
@nomah4 do you plan to work on this?

nomah4:
@amir20 

Yes, I’m planning to work on this.

I’ll need some time to check what already exists, because I haven’t seen the current host-level grouping yet.

nomah4:
Screenshot 2569-04-25 at 13 26 05
Screenshot 2569-04-25 at 13 26 12

Yes — container grouping already works inside a host, but this PR solves a different problem: grouping hosts themselves, especially for remote-agent deployments.

With agent-based setups, users often manage dozens of servers across environments like Production, Development, and Staging. Without host grouping, the sidebar becomes a flat list of hosts, which is difficult to navigate.

Container groups are:

Host → Containers

while this PR adds:

Group → Hosts

for example:

Production

* app-01
* app-02

Development

* test-01
* test-02

Makefile
```diff
@@ -54,6 +54,10 @@ push: docker
 	@docker tag amir20/dozzle:local amir20/dozzle:local-test
 	@docker push amir20/dozzle:local-test
 
+.PHONY: docker-push-dev
```
amir20: This needs to be removed :)

e2e/visual.spec.ts-snapshots/dark-homepage-1-chromium-linux.png
amir20, line 1: Uhh this screenshot doesn't look right. Why is "running containers" gone?

amir20 (changes requested):
Can you explain why "running containers" section is gone?

amir20:
OK i brought this branch locally. Tested it fine. It does seem like the sidemenu "running containers" label is replaced by host now. Which doesn't make sense because user is already on that host.

Makefile
nomah4: Fix :-)

nomah4:
> OK i brought this branch locally. Tested it fine. It does seem like the sidemenu "running containers" label is replaced by host now. Which doesn't make sense because user is already on that host.

Thanks, good catch.
You're right - replacing _running containers_ with the host name was unnecessary since the user is already inside that host context.
I’ve already fixed this and restored the original _running containers_ label so the navigation stays consistent and less confusing.

amir20:
Nice. Ill bring this locally and test one more time before merging.

amir20:
Closed with https://github.com/amir20/dozzle/pull/4662

amir20:
@nomah4 I'll be releasing a new build later this week. For now you can test `master`.

PR #4655, opened 2026-04-26
perf: initialize agent connections in parallel on startup

nomah4:
## Problem

When Dozzle starts with many remote agents configured, `NewRetriableClientManager` iterates over every endpoint **sequentially**. Each `agent.Host()` call either completes a network round-trip or blocks for the full configured timeout before the loop moves to the next endpoint.

With, say, 10 agents each hitting a 10-second `DeadlineExceeded`, startup takes **100 seconds** instead of 10. The same issue exists in `RetryAndList`, which is called periodically to reconnect failed agents — each retry cycle also stalls for `N × timeout`.

Log evidence from a real deployment (sequential, 10s apart):
```
11:44:37 error fetching host info for agent  proxy-4-1
11:44:48 error fetching host info for agent  proxy-4-3   ← +11s
11:44:48 error fetching host info for agent  proxy-4-4
11:44:58 error fetching host info for agent  proxy-4-6   ← +10s
```

## Fix

Fan out a goroutine per endpoint in both `NewRetriableClientManager` and `RetryAndList`, then collect results once all goroutines finish.

**`NewRetriableClientManager`**
- Spawns one goroutine per `clients` entry and one per `agents` entry.
- A local `sync.Mutex` guards concurrent writes to `clientMap`.
- Results flow through buffered channels; both channels are drained after `wg.Wait()`.

**`RetryAndList`**
- Copies `failedAgents` under a read lock so `List`/`Find` callers are never blocked during network I/O.
- Runs all retries concurrently into a pre-allocated `results` slice (index-safe, no channel needed).
- Acquires the write lock only to merge results and rebuild `failedAgents`, preserving any entries added concurrently.

With the fix, startup time equals the slowest single agent timeout instead of their sum.

### DIFF
--- a/internal/support/docker/retriable_client_manager.go
+++ b/internal/support/docker/retriable_client_manager.go
@@ -29,50 +29,80 @@ type RetriableClientManager struct {
 
 func NewRetriableClientManager(agents []string, timeout time.Duration, certs tls.Certificate, clients ...container_support.ClientService) *RetriableClientManager {
 	clientMap := make(map[string]container_support.ClientService)
-	for _, client := range clients {
-		ctx, cancel := context.WithTimeout(context.Background(), timeout)
-		defer cancel()
-		host, err := client.Host(ctx)
-		if err != nil {
-			log.Warn().Err(err).Str("host", host.Name).Msg("error fetching host info for client")
-			continue
-		}
+	var mapMu sync.Mutex
+	var wg sync.WaitGroup
 
-		if _, ok := clientMap[host.ID]; ok {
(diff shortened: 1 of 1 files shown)


internal/support/docker/retriable_client_manager.go
```diff
@@ -29,50 +29,80 @@ type RetriableClientManager struct {
...
-		host, err := client.Host(ctx)
-		if err != nil {
-			log.Warn().Err(err).Str("host", host.Name).Msg("error fetching host info for client")
-			continue
-		}
+	var mapMu sync.Mutex
+	var wg sync.WaitGroup
+
+	for _, c := range clients {
+		wg.Add(1)
```
amir20, line 36: Could you use wg.Go() instead?

amir20 (commented):
Just one minor comment. I can also fix with Claude.

amir20:
Closing as I merged and improved here https://github.com/amir20/dozzle/pull/4656

PR #4665, opened 2026-04-28
feat(cloud): per-replica cloud connections in swarm mode

amir20:
## Summary

- Each Dozzle process sends `x-instance-id` gRPC metadata (its local Docker host ID) so Dozzle Cloud can keep one connection per node instead of rejecting duplicates by API key. In swarm mode every replica holds its own connection — log streaming and tool dispatch happen on the node that owns the data instead of funneling through one "primary".
- Cloud + notification config broadcasts now persist on receiving replicas (previously only the originating replica wrote `cloud.yml` / `notifications.yml`), so `/api/cloud/config` is consistent across replicas after a refresh.
- Threads `CLOUD_URL` as a Docker build arg for local-Cloud testing, adds `/data` volume guidance to swarm + k8s docs, and updates `examples/setup-swarm.fish` for end-to-end local testing against Doligence.

Pairs with the matching Doligence registry change (multi-connection per API key keyed by `(api_key, instance_id)`).

## Test plan

- [ ] `make test` passes
- [ ] Build with `CLOUD_URL=http://localhost:3000 make docker` and verify the bundle ships that URL (not `cloud.dozzle.dev`)
- [ ] `examples/setup-swarm.fish` brings up a 3-node swarm; link cloud once via the browser; verify all three replicas show linked after browser refresh; check Doligence's `/instances` shows three connections; confirm log streaming load is distributed across replicas (not funneled through one)
- [ ] Standalone server (non-swarm) cloud link still works — single connection, unchanged behavior

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,9 @@ COPY assets ./assets
 COPY locales ./locales
 COPY public ./public
 
+ARG CLOUD_URL
+ENV CLOUD_URL=$CLOUD_URL
+
 # Build assets
 RUN pnpm build
 
--- a/internal/support/cli/agent_command.go
+++ b/internal/support/cli/agent_command.go
@@ -7,7 +7,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
-	"sync"
+	"sync/atomic"
 	"syscall"
 
 	"github.com/amir20/dozzle/internal/agent"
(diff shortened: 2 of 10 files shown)


internal/support/cli/agent_command.go
```diff
@@ -27,6 +29,24 @@ type AgentCmd struct {
...
+	mu          sync.RWMutex
+	cloudConfig *notification.CloudConfig
+	onCloudSet  func()
+}
+
+// CloudConfig returns the agent's currently active cloud config, or nil if
+// none has been pushed by the main server / loaded from disk.
+func (h *persistingNotificationHandler) CloudConfig() *notification.CloudConfig {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
```
amir20: can this be atomic instead?

internal/support/docker/multi_host_service.go
```diff
@@ -357,6 +366,71 @@ func (m *MultiHostService) NotificationHandler() *notification.Manager {
 	return m.notificationManager
 }
 
+// SetCloudNotifyFunc registers a callback the agent server invokes after a
+// peer broadcast so the local cloud client reconnects with the new API key.
+func (m *MultiHostService) SetCloudNotifyFunc(fn func()) {
+	m.cloudNotifyMu.Lock()
+	m.cloudNotifyFn = fn
```
amir20: same here, can this be atomic?

Dockerfile
```diff
@@ -19,6 +19,9 @@ COPY assets ./assets
 COPY locales ./locales
 COPY public ./public
 
+ARG CLOUD_URL
+ENV CLOUD_URL=$CLOUD_URL
```
amir20, line 23: Just confirm this won't break if it is set to nothing?

amir20: Confirmed safe: when `CLOUD_URL` is unset at build time, `ARG CLOUD_URL` resolves to empty, `ENV CLOUD_URL=""` is set, and `vite.config.ts:17` does `process.env.CLOUD_URL || "https://cloud.dozzle.dev"` — empty string is JS-falsy so the default URL is used. Verified the production image still bakes the cloud.dozzle.dev URL when no build-arg is passed.

PR #4673, opened 2026-05-04
fix(docker): use cancellable context in ContainerEvents and close event channel

Srujan-rai:
ContainerEvents was previously called with context.Background(), ignoring the collector's lifecycle context. On host disconnect or forceStop(), the event subscription continued running indefinitely.

This change switches to a cancellable context so the subscription properly terminates with the collector lifecycle.

Also adds `defer close(events)` so the consumer goroutine (`for range events`) exits when the producer returns, preventing a second goroutine leak.


amir20:
Nice find. Merging.

Srujan-rai:
> Nice find. Merging.

Thank you

PR #4674, opened 2026-05-04
fix(security): block 0.0.0.0/8 and broadcast in SSRF webhook guard

Srujan-rai:
Problem
                                                                                                                                                              
  isBlockedIP in the webhook SSRF guard missed two address ranges:
                                                                                                                                                              
  - 0.0.0.0/8 — only the exact address 0.0.0.0 was caught by IsUnspecified(). Addresses like 0.1.2.3 through 0.255.255.255 were allowed through. On Linux, these route to the local host, making them a viable SSRF bypass vector.                                                                                     
  - 255.255.255.255 — limited broadcast address, not covered by any existing check.                                                                           
                                                                                                                                                              
  An attacker controlling a webhook URL (or DNS) could point to http://0.1.2.3/ to reach local services on the Dozzle host.                                   
                                                                                                                                                              
  Fix                                                                                                                                                         
                                                            
  - Added zeroNetV4 (0.0.0.0/8) CIDR check in isBlockedIP for IPv4 addresses                                                                                  
  - Added explicit block for net.IPv4bcast (255.255.255.255)
  - Extended TestIsBlockedIP to cover 0.1.2.3, 0.255.255.255, and 255.255.255.255                                                                             
                                                                                                                                                              
  Scope                                                                                                                                                       
                                                                                                                                                              
 No behavior change for legitimate webhooks. RFC1918 private ranges (192.168.x.x, 10.x.x.x, 172.16-31.x.x) remain allowed.

### DIFF
--- a/internal/notification/dispatcher/webhook.go
+++ b/internal/notification/dispatcher/webhook.go
@@ -26,13 +26,28 @@ import (
 // live on private LANs.
 var errBlockedAddress = errors.New("webhook target resolves to a blocked address range")
 
+// zeroNetV4 covers 0.0.0.0/8 — on Linux these route to the local host.
+var zeroNetV4 = &net.IPNet{IP: net.IP{0, 0, 0, 0}, Mask: net.CIDRMask(8, 32)}
+
 func isBlockedIP(ip net.IP) bool {
-	return ip.IsLoopback() ||
+	if ip.IsLoopback() ||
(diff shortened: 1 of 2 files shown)


amir20:
It will be a week before I release these changes.

Srujan-rai:
> It will be a week before I release these changes.

Thanks for the update! Appreciate the merge looking forward to the release.

PR #4684, opened 2026-05-08
feat: integrate MCP server via Streamable HTTP on existing web server

aaronpowell:
## Summary

Replaces the stdio-based MCP subcommand from #4495 with an HTTP-based MCP endpoint served from the same Dozzle container, as discussed in [#4604](https://github.com/amir20/dozzle/discussions/4604#discussioncomment-16823969).

## Changes

- **Switch SDK**: Replace `mark3labs/mcp-go` with the official `github.com/modelcontextprotocol/go-sdk` (Streamable HTTP transport)
- **Mount at `/api/mcp`**: MCP handler is mounted in the existing chi router, inside the authenticated route group — same auth as the rest of the API
- **Feature flag**: Gated behind `--enable-mcp` / `DOZZLE_ENABLE_MCP` (default: off)
- **Read-only tools**:
  - `list_containers` — List all containers with optional state filter
  - `get_container_logs` — Fetch structured logs with configurable time range
  - `list_hosts` — List connected Docker hosts
  - `get_container_stats` — CPU/memory stats history
- **Remove subcommand**: Deleted the separate `dozzle mcp` subcommand and `mcp_command.go`
- **Tests**: Unit tests for all tool handlers using in-memory MCP transports

## Usage

```bash
# Enable MCP endpoint
docker run ... -e DOZZLE_ENABLE_MCP=true amir20/dozzle

# MCP clients connect to:
# http://your-dozzle:8080/api/mcp
```

## VS Code MCP client config

```json
{
  "servers": {
    "dozzle": {
      "type": "http",
      "url": "http://localhost:8080/api/mcp"
    }
  }
}
```

Example from using it in Copilot Chat in VS Code:

image

For this test, I ran two containers, one which was just `ubuntu:latest` which I did a few `echo` commands, the other is a container running an ASP.NET application that produces some warnings when started.

The prompt I gave it was "are there any warnings on my containers?" in which it ran the tools `list_containers` and then `get_container_logs`.


amir20:
Hey @aaronpowell I am little confused because this is different than what we originally discussed in https://github.com/amir20/dozzle/discussions/4604#discussioncomment-16827157 with @ludekvodicka. 

I thought we said we are not doing MCP because that's an overkill. Instead doing a simple CLI with skill.md. 

Now don't get me wrong, I am not against this because it's originally what I implemented.  And most people are running Dozzle remotely so having a CLI would actually complicate things 

In the original PR, I had done an MCP except after testing it, I found that it would not work on Claude Code. It requires https. After doing some investigation I realized that it would be pretty complicated to support HTTPS because the certs needs to be valid too. 

It seems like there is no easy solution for all agents to work. Most people have Dozzle remotely and at the same time HTTPS is usually required with valid HTTPS. 

I might just be over thinking all of this. So hopefully, you all shed some light if this would bring confusion to the users. 

Also this PR needs documentation at `docs/`.

aaronpowell:
I was going off the approach of "what would MCP look like" in the replies over the "add a set of APIs" because as I pointed out, agentic tools (Copilot, Claude, Cursor, etc.) all have MCP built natively into them so know how to make the request whereas to call an API code has to be generated first (generally Python or Node) which then needs to be executed in a shell. So, depending on how much trust you allow your agent, some of that isn't desirable.

I'm unsure of the requirement on SSL (or SSL that isn't self-signed) as I did the test on HTTP via localhost which _maybe_ the client didn't care about it. It's also been a while since I read the MCP spec and I can't recall what it says about client + Streamable HTTP + SSL.

IMO MCP makes sense if the desire is to support agentic coding tools interacting with dozzel but not other types of clients. If the objective is to have other client (coding agents, custom CLIs, custom apps, etc.) then an API is better because it provides a wider surface.

amir20:
I tend to agree. MCP makes more sense. I was playing around with Claude and I think it's easier to have an HTTP stream than CLIs. CLIs would need to be wrapped around another http call for remote deployments anyway. 

Now back to this PR: 

- It doesn't seem like authentication works. Right? It would just be rejected. 
- There is no documentation 
- handleListContainers aborts on any per-host error (server.go:101-105). ListAllContainers returns []error because partial failures are normal (one host down). Returning early hides every other host's containers. Should log/skip and continue.

amir20 (changes requested):
See comments

aaronpowell:
Comments actioned. I _think_ the auth will work because it's mounted on the `/api` route group but I don't actually use auth in my setup so... 🤫 😂

amir20:
I am not sure how many people will actually use the auth path given that JWT is pretty hard to generate. I wonder if the agent can do it. It seems like the right way would be to do with tokens. But I'd say that's out of scope for this PR. I've had MCP on Dozzle Cloud for a couple weeks now and not many people actually seem to care.  

Anyway, I'll do a through pass and merge.

amir20 (approved):

Issue #4678, opened 2026-05-07
What "type" of regex is supported? negative lookaheads don't work


lingfish:
### 🔍 Check for existing issues

- [x] Completed

### How is Dozzle deployed?

Standalone Deployment

### 📦 Dozzle version

v10.5.1

### ✅ Command used to run Dozzle

Very standard `docker-compose`:

```yaml
  dozzle:
    image: amir20/dozzle:latest
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - dozzle_data:/data
```

### 🐛 Describe the bug / provide steps to reproduce it

I'm trying to use find to exclude log lines using a negative lookahead assertion, eg: `(?!health)`

Browser console reports a HTTP 400.

So I guess I'm asking for either a way to filter out/exclude log lines, or what type of regex Dozzle supports.

(And just as I'm enabling DEBUG logs, I think half the answer appears... unsupported perlre.)

It'd be good to have that implemented, or another find/exclude way.

### 💻 Environment

Don't think they're relevant here.

### 📜 Dozzle logs are required for debugging purposes. You may need to enable debug mode. See https://dozzle.dev/guide/debugging.

```
dozzle-1  | {"level":"debug","version":"v10.5.1","error":"error parsing regexp: invalid or unsupported Perl syntax: `(?!`","pattern":"(?!health)","time":"2026-05-07T04:21:34Z","message":"failed to compile regex"}
```

amir20:
Thanks for the report. Dozzle uses Go's standard `regexp` package, which is RE2-based — it intentionally doesn't support lookaheads, lookbehinds, or backreferences to guarantee linear-time matching and avoid ReDoS. That's why `(?!health)` fails to compile.

Switching engines (e.g. to `regexp2`/PCRE) isn't something I want to take on — every connected client runs the pattern against every log line, and giving up the linear-time guarantee is a real risk. Not worth implementing right now.

lingfish:
Ok, understood. Apologies, the regex support isn't clearly docuemnted, or I couldn't find it.

How do you then feel about an exclusion feature (think `grep -v`)? My use case is that a traefik log for example might have a ton of healthcheck pings, and it'd be good to filter those out on the fly:

```
2026-05-08 16:46:55.836 | INFO     | code.api.main:log_requests:215 - Request: GET http://localhost:8000/health
```

amir20:
[...]

PR #4689, opened 2026-05-11
feat: add inverse/exclude filter to log search

lingfish:
## Summary

Adds a grep `-v` style inverse/exclude filter toggle to the log search panel. When active, logs that match the search pattern are **excluded** instead of included.

This is to address https://github.com/amir20/dozzle/issues/4678.

### Backend (Go)
- `internal/web/logs.go`: `matchesFilter()` accepts `inverse bool` using XOR logic (`inverse != support_web.Search(...)`). Parsed at both `fetchLogsBetweenDates` and `streamLogsForContainers` endpoints via `?inverse=true` query param.
- `internal/cloud/tools_logs.go`: Added `Inverse bool` field to `fetchLogsArgs`.
- `internal/cloud/tools_stream.go`: `matchesFilters()` uses XOR for regex + query matching.

### Frontend (Vue 3)
- `assets/composable/search.ts`: Added `inverseFilter` ref and `toggleInverse()`.
- `assets/components/Search.vue`: Inline toggle button with `mdi:filter-off-outline` / `mdi:filter-outline` icons, red tint (`btn-error`) when active.
- URL params propagate `inverse=true` to SSE stream (`eventStreams.ts`), historical fetch (`historicalLogs.ts`), and download (`downloadUrl.ts`) endpoints.
- `visible.ts` flips client-side `<mark>` post-filter for ComplexLogEntry when inverse is active.

### Tests
- `internal/web/logs_test.go`: 6 subtests for `matchesFilter` inverse behavior.
- `internal/cloud/tools_stream_test.go`: 8 subtests for `matchesFilters` inverse (regex + query).

## Motivation

Users monitoring error logs may want to **exclude** verbose noise (e.g., health check pings) rather than include specific patterns. The inline toggle makes this discoverable without needing to remember a `!` prefix convention.

### DIFF
--- a/assets/components/Search.vue
+++ b/assets/components/Search.vue
@@ -17,6 +17,15 @@
           v-model="searchQueryFilter"
           @keyup.esc="resetSearch()"
         />
+        <button
+          class="btn btn-circle btn-xs"
+          :class="inverseFilter ? 'btn-error' : 'btn-ghost'"
+          @click="toggleInverse()"
+          :title="inverseFilter ? 'Exclude matching (inverse)' : 'Include matching (normal)'"
+        >
+          <mdi:filter-off-outline v-if="inverseFilter" />
+          <mdi:filter-outline v-else />
+        </button>
         <a class="btn btn-circle btn-xs" @click="resetSearch()"> <mdi:close /></a>
       </div>
     </div>
@@ -26,7 +35,7 @@
 <script lang="ts" setup>
 const input = ref<HTMLInputElement>();
 const container = ref<HTMLDivElement>();
-const { searchQueryFilter, showSearch, resetSearch, isValidQuery } = useSearchFilter();
+const { searchQueryFilter, showSearch, resetSearch, isValidQuery, inverseFilter, toggleInverse } = useSearchFilter();
 
 const { style } = useDraggable(container);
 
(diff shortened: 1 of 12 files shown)


assets/components/Search.vue
```diff
@@ -17,6 +17,15 @@
           v-model="searchQueryFilter"
           @keyup.esc="resetSearch()"
         />
+        <button
+          class="btn btn-circle btn-xs"
+          :class="inverseFilter ? 'btn-error' : 'btn-ghost'"
+          @click="toggleInverse()"
+          :title="inverseFilter ? 'Exclude matching (inverse)' : 'Include matching (normal)'"
```
amir20, line 24: This should be added to locale files. If it's easy I'll do it.

amir20:
Fixed the comments in https://github.com/amir20/dozzle/pull/4690

amir20:
Merged.

lingfish:
@amir20 does GitHub not let you @ tag me? Have I mangled my profile/privacy or something? I just noticed it missing in these PRs and the release notes.

amir20:
I merged locally by addressing some of Claude feedback. I am not sure how PRs work and if you still get tagged. 🤷🏼

PR #4698, opened 2026-05-13
fix: update key binds

### LINKED ISSUE #4693: Clear Log keyboard shortcut conflicts with Firefox

Thanks for making this!  I love that it's possible to clear the log entries from the screen without destroying logs, and often use this feature to make it easier to review logs when troubleshooting containers.  Unfortunately, I cannot use the keyboard shortcut for this feature because it conflicts with my browser.

The `Ctrl(/Meta) + Shift + K` shortcut conflicts with Firefox's [Open Web Console](https://firefox-source-docs.mozilla.org/devtools-user/keyboard_shortcuts/index.html).  Unfortunately, both your app and Firefox seem to have this hard-coded and therefore I cannot reassign either.  Firefox sees the event first and traps it, so the event never makes it to your app's code.  Disabling the web console isn't an option either.  

Would you please consider making the shortcuts adjustable by user preference, or simply moving the shortcut to something that doesn't conflict with browser behavior?

In Linux, `Ctrl+L` is command some use to clear the screen.  Unfortunately, some browsers (Firefox and Chrome, at least) use this to focus on the address bar.  However, `Ctrl + Shift + L` doesn't *seem* to be a commonly used keyboard in browsers.  Perhaps this would be a viable option?  This is what I would select if the keybind was configurable.  

_No response_

dever23b:
### Description
* Fixes [#4693](https://github.com/amir20/dozzle/issues/4693)

### Notes
* I have extremely limited experience with Go, so forgive me if this seems more obvious to a more experienced developer.  I was not able to build this package using only the instructions at [Building](https://github.com/amir20/dozzle#building):
  > missing go.sum entry for module providing package google.golang.org/grpc/cmd/protoc-gen-go-grpc; to add:
  >        go mod download google.golang.org/grpc/cmd/protoc-gen-go-grpc
* I first had to use command `go mod tidy` prior to `go install tool`.  I noticed that this changed a few other files in the repository, and I don't understand the changes well enough to know whether this step may cause other problems.  The workaround allowed me to build and test this PR, though.  You may want to consider reviewing your build instructions.
* I'm not sure if this works on other browsers, but during my testing on Firefox I discovered that none of your defined keystroke event handlers intended to catch Shift-modified keystrokes were working.  I updated any applicable `onKeyStroke` invocations.

### Changes
* `assets/components/ContainerViewer/ContainerActionsToolbar.vue`
  > * Updated UI to reflect new key bind
  > * Updated `onKeyStroke` parameters to detect presence of Shift-modified keystrokes

* `assets/components/LogViewer/MultiContainerActionToolbar.vue`
  > * Updated UI to reflect new key bind

* `assets/components/LogViewer/ViewerWithSource.vue`
  > * Updated logic to detect new key bind

### DIFF
--- a/assets/components/ContainerViewer/ContainerActionsToolbar.vue
+++ b/assets/components/ContainerViewer/ContainerActionsToolbar.vue
@@ -18,7 +18,7 @@
       <li v-if="!historical">
         <a @click="clear()">
           <octicon:trash-24 /> {{ $t("toolbar.clear") }}
-          <KeyShortcut char="k" :modifiers="['shift', 'meta']" />
+          <KeyShortcut char="l" :modifiers="['shift', 'meta']" />
         </a>
       </li>
       <li v-if="hasComplexLogs">
@@ -299,7 +299,7 @@ async function copyLogs() {
   await navigator.clipboard.write([new ClipboardItem({ "text/plain": blobPromise })]);
 }
 
-onKeyStroke("f", (e) => {
+onKeyStroke(["f", "F"], (e) => {
   if (hasComplexLogs.value) {
     if ((e.ctrlKey || e.metaKey) && e.shiftKey) {
       showDrawer(LogAnalytics, { container }, "lg");
@@ -308,14 +308,14 @@ onKeyStroke("f", (e) => {
   }
 });
 if (enableShell) {
-  onKeyStroke("a", (e) => {
+  onKeyStroke(["a", "A"], (e) => {
     if ((e.ctrlKey || e.metaKey) && e.shiftKey) {
       showDrawer(Terminal, { container, action: "attach" }, "lg");
       e.preventDefault();
     }
   });
 
-  onKeyStroke("e", (e) => {
+  onKeyStroke(["e", "E"], (e) => {
     if ((e.ctrlKey || e.metaKey) && e.shiftKey) {
       showDrawer(Terminal, { container, action: "exec" }, "lg");
       e.preventDefault();
--- a/assets/components/LogViewer/MultiContainerActionToolbar.vue
+++ b/assets/components/LogViewer/MultiContainerActionToolbar.vue
@@ -12,7 +12,7 @@
       <li>
         <a @click="clear()">
           <octicon:trash-24 /> {{ $t("toolbar.clear") }}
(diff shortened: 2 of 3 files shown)


amir20:
@dever23b I am trying to understand why you added the uppercase variations. It currently works for me without those changes on Chrome based and Firefox on Mac. Is it not working for you? For example `cmd + shift + f` works just fine.

dever23b:
> @dever23b I am trying to understand why you added the uppercase variations. It currently works for me without those changes on Chrome based and Firefox on Mac. Is it not working for you? For example `cmd + shift + f` works just fine.

Perhaps it's an OS/browser implementation then?  For me, on Windows/Firefox, none of the original keystroke event handlers were firing whenever I held down `Shift`.

To test, I wrote:
```typescript
onKeyStroke("l", (e) => void console.log("l key pressed %O", {
  shift: e.shiftKey,
  meta: e.metaKey,
  ctrl: e.ctrlKey 
}))
```

I observed that the event did not fire at all if I pressed `Shift + l`.  I've never used Vue before, so I had to do some research on the [function implementation](https://vueuse.org/core/onKeyStroke/#onkeystroke).  Though the documentation doesn't mention it, all of the examples used an array containing both cases of the letter being targeted  When I converted the `key` to an array containing the lowercase and uppercase variants to match the documentation examples, it started working for me.

dever23b:
I did some more research.  Obviously, Mozilla's [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) supports my findings, but I looked elsewhere to see whether [other sources](https://www.w3schools.com/jsref/event_key_key.asp) agreed.  This seems to be a standard behavior of the `KeyboardEvent.key` property, which I believe is what Vue's `onKeyStroke` may be evaluating.

When `Shift` is held down, the returned `event.key` property should return the uppercase variant(/symbol) of whatever key was pressed.  Accordingly, if there is no listener for the uppercase variant, the handler does not fire.  This was my experience in testing.

amir20:
Hmmm interesting. Seems to be an OS thing. I don't have another one handy. 

I wonder if it has been broken forever and no one ever said anything 🤷🏼

i'll merge then.

dever23b:
Thanks!

PR #4700, opened 2026-05-14
fix: pin @codemirror/view to single version via pnpm override

amir20:
## Summary
- Adds a pnpm override forcing `@codemirror/view` to `^6.43.0` so transitive deps (`@codemirror/autocomplete`, `@codemirror/commands`, etc.) dedupe onto the same version as the direct dep
- Without this, two copies (`6.42.1` and `6.43.0`) coexisted in the lockfile, producing `vue-tsc` errors about mismatched private types on `EditorView.dispatchTransactions`
- Unblocks the renovate PR #4699

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm why @codemirror/view` reports a single version


amir20:
Superseded — fixed directly on the renovate branch by regenerating pnpm-lock.yaml, no override needed.

PR #4710, opened 2026-05-18
feat(welcome): reframe cloud setup as triage with selectable default signals

amir20:
## Summary

Reframes the post-link welcome modal from "we created a default alert for you" to "Cloud is an intelligent triage layer — here's what it listens for; you decide what we hear."

### What changes

- **Copy.** Title, body, and footer rewritten around the triage framing. Footer reinforces consent: *"Cloud never receives logs you haven't subscribed it to."*
- **One rule → four-row checklist.** Each row is a checkbox + label + description. On submit, one POST per checked row to `/api/notifications/rules`, all on the cloud dispatcher.

  | Default | Signal | Event expression |
  |:---:|---|---|
  | on  | Container exited with an error | `name == "die" && attributes["exitCode"] != "0"` |
  | on  | Container became unhealthy | `name == "health_status" && attributes["healthStatus"] == "unhealthy"` |
  | on  | Container was killed by the kernel (OOM) | `name == "oom"` |
  | off | Container restarted | `name == "restart"` |

- **No highlight redirect.** With N rules, picking one to highlight is arbitrary — land on `/notifications` plain instead.
- **i18n.** New keys added to `locales/en.yml`; 16 non-English locales receive English fallback for the new keys and will be translated through the normal localization flow.
- **Test.** `WelcomeModal.spec.ts` now asserts 3 POSTs by default with the right `eventExpression` per signal.

### Companion PR

The OOM row depends on #4709 (`feat: allow oom and kill events through notification listener`) to be functional end-to-end. Without it the rule still saves cleanly — it just matches nothing until the listener whitelist lands.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm test` (12 passed)
- [x] `make test` (all packages green)
- [ ] Manual: link a fresh instance and confirm all 3 default rules appear on `/notifications` after the modal flow


amir20:
Addressed in `a517ad56`. Walking through the consolidated feedback:

**Applied:**
- `Promise.all` for rule POSTs (was `for...of` sequential).
- Dropped the unreachable `chosen.length === 0` early-return.
- Rule names are now stable English (`ruleName` on each `SignalDef`) so the user's `/notifications` page doesn't show garbled names if they switch locale.
- `flushPromises()` in the spec instead of the magic 5-iter `setTimeout(0)` loop.
- `step2-subtitle`, `checklist-*`, and `default-alert-name` orphan keys removed across all 17 locales.
- `title` and `step2-title` were identical in every locale — `step2-title` is now "Pick the signals Cloud should triage." (translated in 16 non-English locales).
- Step1 subtitle softened: "is already turned on" → "is pre-selected" (rules aren't created until the user submits step2).
- Pirate locale's plain-English `step2-body`, `footer`, `signals.*`, and `create-alerts` are now pirate-voiced.
- On the `onNotificationsPage` fast-path, **skip** no longer silently calls `createDefaultAlerts()` — it just dismisses. Submit still creates defaults (the user confirmed intent). This matches the consent framing the PR is building.

**Pushed back:**
- *"`selectedSignals` is empty on `onNotificationsPage` fast-path → no rules created"* — not a bug. `open()` seeds `selectedSignals` from `signals.value.filter(s => s.defaultOn)` before either `submitFeedback` or `skipFeedback` can run. The submit-fast-path correctly creates the 3 defaults.
- *"`signals` doesn't need `computed`"* — keeping it computed. It depends on `t()`, which is reactive to locale changes; a plain `const` evaluated at setup time would render stale labels after a locale switch.

**Acknowledged, not fixed:**
- Partial failure with `Promise.all` is still all-or-nothing-from-the-user's-perspective but earlier-resolved POSTs aren't rolled back. Added an inline comment noting this is acceptable for a welcome modal — the user can edit/delete rules from `/notifications`.

CI: `pnpm typecheck`, `pnpm test`, and `make test` all green locally.

PR #4722, opened 2026-05-20
docs: update podman.md guide with quadlet info

LeeNX:
Updated details include
 - compatible notes for podman vs docker
 - examples for quadlet's
 - info on healthchecks

PR #4723, opened 2026-05-20
docs: fix incorrect podman.md examples

amir20:
## Summary

Fixes several incorrect or broken examples in the Podman guide that landed in #4722.

- Agent Quadlet had `Entrypoint=agent`, but the image's entrypoint is `/dozzle` and `agent` is a subcommand. Moved to `Exec=agent`.
- Replaced bogus `DOZZLE_AGENT_FILTER` env var with the actual `DOZZLE_REMOTE_AGENT`.
- Dropped `WantedBy=multi-user.target` from user units (system-only target, doesn't apply to `systemctl --user`).
- `HealthCmd=` takes a plain command, not the Docker `CMD [...]` JSON form. Removed the stray `CMD` prefix.
- Switched Quadlet examples to canonical keys: `PublishPort=` / `Volume=` / `Environment=` instead of compose-style `Ports=` / `Volumes=`.
- Rewrote the rootless memory stats FAQ. The previous fix (`DefaultCgroupsMode=unified` in `/etc/systemd/system.conf`) isn't a real directive; replaced with the correct `Delegate=` drop-in for `user@.service`.
- Collapsed the per-user-namespace and Quadlet sections (near-duplicates) and dropped the "Web GUI" naming in favor of "Standalone".

## Test plan

- [ ] Render docs site locally and skim the Podman page
- [ ] Confirm the Quadlet snippets parse with `/usr/libexec/podman/quadlet -dryrun` on a Podman host

### DIFF
--- a/docs/guide/podman.md
+++ b/docs/guide/podman.md
@@ -4,31 +4,30 @@ title: Podman
 
 # Podman
 
-Dozzle supports Podman through its Docker-compatible socket interface. However, Podman is not 100% compatible with Docker, key differences include memory stats reporting (especially in rootless/Quadlet deployments) and the lack of automatic engine-id generation. This guide covers both Web GUI mode (standalone, local monitoring) and Agent mode (remote monitoring via a central Dozzle server), with notes on these compatibility considerations.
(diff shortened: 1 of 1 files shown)


docs/guide/podman.md
```diff
@@ -252,7 +203,7 @@ Or with environment variables:
 ```bash
 podman run -d \
   --name dozzle \
-  -e DOZZLE_AGENT_FILTER="host1.example.com:7007,host2.example.com:7007" \
+  -e DOZZLE_REMOTE_AGENT="host1.example.com:7007,host2.example.com:7007" \
```
amir20, line 206: This was wrong

PR #4729, opened 2026-05-22
feat(cloud): per-container log filtering via dev.dozzle.cloud.min_level label

amir20:
## Summary

Adds a `dev.dozzle.cloud.min_level` container label that controls what gets forwarded to Dozzle Cloud. Closes #4727.

| Value | Effect |
| --- | --- |
| _(unset)_ | All log lines forwarded. Default. |
| `disabled` | Container skipped entirely. No reader started. |
| `trace` / `debug` / `info` / `warn` / `error` / `fatal` | Only lines at that level or higher forwarded. Lines without a detected level always pass. |

Filtering happens on the host inside `logStreamer` before batching, so dropped lines never touch the network or count against the user's plan. Local log viewing is unaffected.

## Why

From #4727: noisy containers (e.g. zigbee2mqtt publishing MQTT events) flood the cloud index with info-level chatter that has no diagnostic value. A label-based control keeps config next to the container, avoids UI surface area, and matches the existing `dev.dozzle.name` / `dev.dozzle.group` convention.

## Implementation

- `internal/cloud/log_streamer.go`
  - New `parseMinLevel()` reads the label, returns `(rank, disabled)`
  - `startReader` short-circuits when `disabled`
  - `runReader` drops events where `cloudLevelRank[ev.Level] > 0 && < minRank`. `unknown` levels rank 0 and always pass.
- `docs/guide/dozzle-cloud.md` — new "Controlling What Gets Forwarded" section with the table and a compose example.

## Test plan

- [x] `go test -race ./internal/cloud/` passes
- [x] `TestParseMinLevel` covers empty / `disabled` / canonical levels / garbage
- [x] `TestLogStreamer_LabelDisabledSkipsContainer` verifies `StreamLogs` is never called
- [x] `TestLogStreamer_LabelMinLevelFiltersBelow` verifies debug/info dropped, warn/error/unknown pass
- [ ] Manual: link a local instance to Cloud, add `dev.dozzle.cloud.min_level=warn` to a noisy container, confirm cloud index only receives warn/error


amir20:
Addressed in 9cd4bda8:

- invalid label values (typos like `warning`) now log an error and stream everything, instead of silently no-opping. `parseMinLevel` returns a `valid` flag, `startReader` logs when it's false.
- replaced the 200ms sleep in `TestLogStreamer_LabelDisabledSkipsContainer` with a canary reader on a second host as a deterministic sync point.
- doc table now notes `trace` == unset, invalid-value behavior, and that the label is read at reader start (restart to apply).

PR #4733, opened 2026-05-23
fix(level-guesser): recognize Zigbee2MQTT-style log levels

amir20:
## Summary

- Detect ` <level>:` (e.g. `[ts] info: msg`) and `:<level> ` (e.g. `Zigbee2MQTT:info  msg`) so these containers get a real level instead of `unknown`.
- This is what was blocking `dev.dozzle.cloud.min_level` from filtering noisy Zigbee2MQTT output (see #4727 discussion). Once levels are detected, the cloud filter works as designed.
- Collapsed the 6-field `levelPatterns` struct (30 compiled regexes) into one combined regex per level (6 total). Same semantics, single `MatchString` per group.
- Removed the unreachable `map[string]any` / `map[string]string` panic branches in `guessLogLevel`.

## Test plan

- [x] `go test ./internal/container/ -run TestGuessLogLevel`
- [x] `go test ./...`
- [ ] Pull `amir20/dozzle:pr-XXXX` and confirm a Zigbee2MQTT container's logs now show levels in the UI

### DIFF
--- a/internal/container/level_guesser.go
+++ b/internal/container/level_guesser.go
@@ -23,15 +23,20 @@ var logLevels = [][]string{
 // aliasToCanonical maps every alias to its canonical level name.
 var aliasToCanonical = map[string]string{}
 
-type levelPatterns struct {
-	plain     *regexp.Regexp // e.g. ^error[^a-z]
-	bracket   *regexp.Regexp // e.g. [ error ]
-	separator *regexp.Regexp // e.g. " error/"
-	quoted    *regexp.Regexp // e.g. "ERROR"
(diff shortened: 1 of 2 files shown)


amir20:
Addressed in c7ef0b11:

- colon-prefix now requires a word char before the colon (`\w:<level>\s`), so it only fires on `tag:level ` and never on `:level` floating in a URL.
- added negative tests for level words in prose and URLs (info in "information", `/info` path, etc).
- ToUpper on the group is intentional. metachars in `(?:error|err)` are unaffected, only the letters flip.

PR #4791, opened 2026-06-16
fix(ui): highlighted URLs staying clickable

PattyRich:
When searching text with the search functionality links break when they are highlighted. This fix allows links to still be clickable even if they have search targets in them.

I also opened up localhost links to be clickable as I think people using dozzle with a dev env would find this useful. Not sure about security of that so can remove If needed.


amir20:
👍🏼

PR #4811, opened 2026-06-30
feat(k8s): group containers by owner chain

chirino:
## Summary
Group Kubernetes containers by every owner in their controller chain so Deployment-owned pods are visible under both their ReplicaSet and Deployment. This makes the Kubernetes menu line up with the workload users actually operate while preserving lower-level owner groupings.

## Changes
- Walk Kubernetes owner references best-effort with dynamic client lookups, stopping cleanly on RBAC errors or Node owners.
- Add stable synthetic `@k8s` owner labels and URL-safe `~` owner keys for namespace/kind/name-safe grouping.
- Update K8s owner routes, menu links, log streams, RBAC docs, and focused backend/frontend tests.


chirino:
Screenshot showing the new groupings:
image

The nice things is can get a stable URL to all the log files for a given deployment.

amir20:
I honestly don't use k8s too much. if you think this looks good i can merge.

amir20:
Two things before this is good to merge.

Blocker: the owner cache poisons negatives on transient errors. `fetchOwnerReferences` returns `false` for every error including context cancellation, deadline, and throttling, and `lookupOwnerReferences` caches that with no TTL. When a request gets cancelled mid-flight (user navigates away while the RS lookup is still running), the owner UID gets cached as not-found permanently and the Deployment drops out of the menu until the process restarts. Only cache real negatives (403/404), or bail when `ctx.Err() != nil`, so transient failures retry next time.

Minor: in `getK8sOwnerRefs` the display kind uses `type` first (`labelValue("type") ?? labelValue("kind")`). For a CRD owner `type` is the disambiguated key like `argoproj.io~v1alpha1~Rollout`, so the menu renders `argoproj.io~v1alpha1~Rollout/api` instead of `Rollout/api`. Use `kind` for the label and keep `type`/`key` for identity.

Rest of what I found (both indexed label sets emitted but only the `@k8s.` one is read, unbounded owner cache growth, synthetic labels leaking into the UI label list) is non-blocking cleanup. Can file separately if useful.

PR #4812, opened 2026-07-01
fix(k8s): trim owner labels, bound owner cache, refresh discovery

amir20:
Follow-up cleanup on the owner-chain grouping from #4811.

## Changes
- Drop the duplicated legacy `k8s.owner.*` indexed labels. Only `@k8s.owner.*` synthetic labels plus top-level `owner.kind/name/key` are emitted now, halving the per-container label payload that ships over SSE/gRPC. The frontend already reads the synthetic set first, so the legacy indexed fallback was dead weight.
- Add a TTL (5m) and a hard size cap to the owner lookup cache. It could previously grow unbounded under ReplicaSet churn, and Forbidden/NotFound negatives were cached forever, so owner chains never recovered after RBAC was granted without a restart. Entries now expire and get swept.
- Reset the discovery REST mapper once and retry on mapping failure, so CRDs registered after startup (Argo Rollouts, etc.) resolve instead of staying unmapped until restart.

## Notes
Owner lookups in the `ContainerEvents` watch loop are still synchronous, but the cache keys on owner UID so shared controllers (ReplicaSet/Deployment) resolve once per rollout. Leaving that on the hot path since moving it off would risk event reordering for little gain.

Tests updated and passing (`go test -race ./internal/k8s`, frontend `k8s.spec.ts`, typecheck).

### DIFF
--- a/assets/stores/k8s.ts
+++ b/assets/stores/k8s.ts
@@ -53,19 +53,16 @@ export function ownerMembershipLabel(key: string) {
 }
 
 export function getK8sOwnerRefs(container: Container): K8sOwnerRef[] {
-  const count = Number(container.labels["@k8s.owner.count"] ?? container.labels["k8s.owner.count"] ?? 0);
+  const count = Number(container.labels["@k8s.owner.count"] ?? 0);
   if (count > 0) {
     const owners: K8sOwnerRef[] = [];
     for (let i = 0; i < count; i++) {
-      const syntheticPrefix = `@k8s.owner.${i}.`;
-      const legacyPrefix = `k8s.owner.${i}.`;
-      const labelValue = (key: string) =>
-        container.labels[`${syntheticPrefix}${key}`] ?? container.labels[`${legacyPrefix}${key}`];
-      const kind = labelValue("kind");
-      const name = labelValue("name");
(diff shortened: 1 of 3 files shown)


amir20:
Addressed in 0c3766b:

- Discovery reset is now throttled to once/min, so an unmappable owner kind can't hammer the discovery endpoint on every pod event.
- pruneOwnerCache evicts down to 3/4 cap instead of clearing wholesale, no more refetch-all thrash past the cap.
- Added tests for TTL expiry refetch, graceful eviction, and the throttled reset.

On caching negative mapping results: skipped it on purpose so a CRD installed after startup still gets picked up within a minute rather than being stuck for the full TTL. The throttle keeps the cost bounded either way.

Fixed the "reset once" comment wording too.

amir20:
Addressed in 5f6ec34:

- pruneOwnerCache now gates eviction on \`> ownerCacheMaxSize\`, so the cache actually grows to max before dropping to evictTo. The hysteresis the comment describes now happens, and ownerCacheMaxSize is used instead of being dead. Fixed the comment wording too.

Existing eviction test still covers it (inserts maxSize+100, asserts trim down to evictTo).

PR #4840, opened 2026-07-15
fix(utils): prevent panic when pushing to a zero-size ring buffer

TowyTowy:
`NewRingBuffer(0)` returns a buffer whose first `Push` panics with an index-out-of-range (and would take a modulo by zero on `r.Size`). It's reachable from the logs endpoint: `min` is validated only as `minimum < 0`, so `?min=0` passes and builds `NewRingBuffer(0)`, then the first matching log line crashes the request handler.

`Push` now returns early when capacity is non-positive, so a zero-size buffer simply holds nothing. `Data`/`Len`/`Clear` were already safe at size 0. Added `TestRingBuffer_ZeroSize`, which panics before the fix and passes after.

Verified locally: `go test ./internal/utils/` (incl. `-race`), `gofmt -l`, and `go vet ./internal/utils/` all clean.

Prepared with AI assistance (Claude) and reviewed/verified by me.

### DIFF
--- a/internal/utils/ring_buffer_test.go
+++ b/internal/utils/ring_buffer_test.go
@@ -44,6 +44,24 @@ func TestRingBuffer_MarshalJSON(t *testing.T) {
 	}
 }
 
+func TestRingBuffer_ZeroSize(t *testing.T) {
+	// A zero-capacity buffer must hold nothing rather than panic. This is
+	// reachable from the logs endpoint (?min=0), which builds NewRingBuffer(0).
+	rb := NewRingBuffer[int](0)
+
+	rb.Push(1)
+	rb.Push(2)
+
+	if rb.Len() != 0 {
+		t.Errorf("Expected len to be 0, got %d", rb.Len())
+	}
+
+	data := rb.Data()
+	if len(data) != 0 {
+		t.Errorf("Expected data to be empty, got %v", data)
+	}
+}
+
 func TestRingBuffer_Clear(t *testing.T) {
 	rb := NewRingBuffer[int](3)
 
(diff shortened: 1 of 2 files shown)


amir20:
Was there a place that RingBuffer of size 0 was being used. Otherwise this shouldn't happen

amir20 (changes requested):
See comment

internal/utils/ring_buffer_test.go
```diff
@@ -44,6 +44,24 @@ func TestRingBuffer_MarshalJSON(t *testing.T) {
 	}
 }
 
+func TestRingBuffer_ZeroSize(t *testing.T) {
+	// A zero-capacity buffer must hold nothing rather than panic. This is
+	// reachable from the logs endpoint (?min=0), which builds NewRingBuffer(0).
+	rb := NewRingBuffer[int](0)
```
amir20, line 50: Does this actually happen any where?

TowyTowy:
Good question, and yes it is reachable from the logs endpoint. In `internal/web/logs.go` the `min` param is validated only as:

```go
if minimum < 0 || minimum > buffer.Size {
    http.Error(w, "minimum must be between 0 and buffer size", http.StatusBadRequest)
    return
}
buffer = utils.NewRingBuffer[*container.LogEvent](minimum)
```

so `?min=0` passes the check (0 is not `< 0`, and `0 > 500` is false) and constructs `NewRingBuffer(0)`. The first matching log line then hits `Push` on the zero-size buffer and panics with index-out-of-range (and would be a modulo-by-zero on `r.Size`).

The frontend never sends `min=0`, so normal UI use never triggers it, but the endpoint accepts it, so any client requesting `?min=0` crashes the handler. Since a zero-capacity buffer is a valid "keep no backlog" state, having `Push` treat it as a no-op seemed like the least surprising fix and keeps the utility safe for any caller.

That said, if you would rather reject `min == 0` at the endpoint instead (e.g. `minimum < 1`), I am glad to switch the PR to that. Whichever you prefer.

PR #4846, opened 2026-07-18
fix: deliver last log line when it has no trailing newline

TowyTowy:
When a container's final write has no trailing newline, the line arrives together with EOF. The docker and k8s log readers returned an empty string on any read error, so that last line was silently dropped from historical fetches and downloads. `docker logs` shows it.

This returns the partial message along with the error instead. The event generator already emits a non-empty message before handling the reader error, so no consumer changes are needed. Added regression tests for both readers (non-tty, tty, and k8s), plus baseline tests for multiplexed frames and >16KB continued frames.

Repro: `docker run --name t alpine sh -c 'printf "done"'`, then fetch/download logs in Dozzle - "done" is missing without this fix.

Disclosure: this fix was developed with AI assistance (Claude); I reviewed and verified the change and tests.

### DIFF
--- a/internal/docker/log_reader.go
+++ b/internal/docker/log_reader.go
@@ -41,9 +41,6 @@ func NewLogReader(r io.Reader, tty bool) *LogReader {
 
 func (d *LogReader) Read() (string, container.StdType, error) {
 	message, stdType, err := d.readEvent()
-	if err != nil {
-		return "", 0, err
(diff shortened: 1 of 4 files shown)


amir20:
I'll merge this but how did you discover this? I don't know if any logging framework that doesn't append line return.

PR #4856, opened 2026-07-22
feat: UI Tweaks on sidebar, search, and settings menu

LBYPatrick:
I've been using Dozzle for over a year and absolutely love it! This batch of frontend tweaks grew out of using it extensively.

## Log viewer / search

- [feat] Search bar no longer covers the actions menu (the menu one sets STDOUT / STDERR, download logs, etc.) button and RAM. Its default position used to sit right on top of both items, preventing user to see RAM or open the actions menu. It now opens below the top bar and, from that default position, animates aside when the actions menu opens. It is also draggable, with press-and-hold on the icon to reset its position.
- [fix] Actions menu works on touch now. It was hover-only and auto-hid on tap-focus, so it never opened on mobile. It opens on tap now and only uses hover where the device supports it.
- [feat] CPU usage toggle. Brings back the older per-core (Linux style) display as an opt-in, alongside the current whole-CPU % default. Stored per profile.
- [feat] "Go to top" and "go to bottom" buttons combined into one rounded widget (tonal weight for up, contained for down). Go to top jumps straight to the container's first lines (fetching just the oldest window, not everything in between); go to bottom returns to the live tail.
- [fix] Fix the command palette drop shadow, which was clipped at the bottom edge.

## Sidebar

- [feat] The edge handle for show / hiding sidebar now has a hotkey `Cmd+Ctrl+S`, and the handle itself now features a label and a hotkey icon on hover.
- [feat] Added "Reset sidebar width" action in the actions menu.
- [feat] "Search containers" bar at the sidebar foot on non-dashboard pages.
- [fix]: Sidebar width did not have consistent width across initial page load and expand / collapse. Does not affect manual resize. 

## Settings

- [feat] Optional fullscreen settings popup (default off, flagged experimental). The `/settings` route is unchanged, and the popup reuses the same panel.
- [feat] Settings is now an item in the actions menu. Combined with the popup, that lets you change preferences without leaving the current page.
- [feat] On non home routes, the 'Hosts and Containers' text is replaced by a smaller version of 'Search Containers' widget previously found in home route at the top of the page. Hotkey also works.

## Other Miscellaneous Items

- [fix] Ghost button hover ring is theme-aware (light on dark, dark on light). It used to be a fixed near-black that failed contrast in dark mode.
- [fix] Level chip labels stay legible on the light theme. Dark labels on the saturated chip backgrounds (info, debug, error, fatal, trace) failed contrast, so they are white there. Warn keeps its dark label on the lighter orange.
- [fix] Mobile: padded top nav with the container status bar locked to it.
- [fix] Reserve the document scrollbar gutter so the sidebar no longer flicks when a page scrollbar appears as logs load.
- [fix] Now clipboard also works under non-secure contexts. This should reduce the amount of 'Clipboard not available' cases.
- [feat]: Docker tag is copy-able by clicking on it. 

## Demo

For the new features and fixes (Some not demonstrated because of GitHub video size limit):

https://github.com/user-attachments/assets/b86f6c26-f08d-420a-8aa8-68c25cfb4f48

For reproducing some issues mentioned in this PR:

https://github.com/user-attachments/assets/2f338f53-baa5-4d55-8f23-e31594f51509

https://github.com/user-attachments/assets/4cc96742-06a3-4d69-9de6-0f022770885d

Again, thank you so much for creating and maintaining this project!


LBYPatrick:
I am fixing the e2e tests, and will convert this PR out of draft when it's good to go. Hang tight!

amir20:
Hi @LBYPatrick , I appreciate the time you put here. This PR is very big. Some of the changes makes sense and some of it I am not sure about. 

Instead of sending a big PR like this, I ask for you to split it up in to multiple feature PR so that we can talk about each specific one. 

I am not going to look at the code.  Here is some feedback: 

> [feat] Search bar no longer covers the actions menu (the menu one sets STDOUT / STDERR, download logs, etc.) button and RAM. Its default position used to sit right on top of both items, preventing user to see RAM or open the actions menu. It now opens below the top bar and, from that default position, animates aside when the actions menu opens. It is also draggable, with press-and-hold on the icon to reset its position.

This was always draggle.  And now it blocks the progress status instead. I am not sure which is better. But that's why I made it draggable. 

Create a PR just for this to discuss. 

> [fix] Actions menu works on touch now. It was hover-only and auto-hid on tap-focus, so it never opened on mobile. It opens on tap now and only uses hover where the device supports it.

We can put all the fixes related to mobile in one PR. 

> [feat] CPU usage toggle. Brings back the older per-core (Linux style) display as an opt-in, alongside the current whole-CPU % default. Stored per profile.

This needs to also take affect on the alerts. It should be it's own PR. I am actually unsure if this would be valuable to add. No one has asked for it. 

> [feat] "Go to top" and "go to bottom" buttons combined into one rounded widget (tonal weight for up, contained for down). Go to top jumps straight to the container's first lines (fetching just the oldest window, not everything in between); go to bottom returns to the live tail.

I like this. I didn't see it in the video. I'll have to play with it to see if I like it. Send it as a PR. 

> Sidebar

All the sidebar ones can be separate. 

On settings: 

> [feat] Optional fullscreen settings popup (default off, flagged experimental). The /settings route is unchanged, and the popup reuses the same panel.

Why? How often are people going to settings page that this needs to be triggered from any page? Could we discuss that? I'd want to know what needs to be changed often. 

>[fix] Ghost button hover ring is theme-aware (light on dark, dark on light). It used to be a fixed near-black that failed contrast in dark mode.
>[[fix] Level chip labels stay legible on the light theme. Dark labels on the saturated chip backgrounds (info, debug, >>[[error, fatal, trace) failed contrast, so they are white there. Warn keeps its dark label on the lighter orange.
>[[fix] Mobile: padded top nav with the container status bar locked to it.
>[[fix] Reserve the document scrollbar gutter so the sidebar no longer flicks when a page scrollbar appears as logs load.
>[[fix] Now clipboard also works under non-secure contexts. This should reduce the amount of 'Clipboard not available' cases.
>[[feat]: Docker tag is copy-able by clicking on it.

These are a little improvements that be combined into one. I'd actually start with these in a separate PR. 

I have to look at the implementation to make sure nothing was introduced.

amir20:
I played around with the code. Some of it is good. I don't agree that the drop down should have any items not related to that specific container. That dropdown is about that page. Having things like reset width or show settings are completely irrelevant to everything else.

LBYPatrick:
Thanks for the feedback Amir! I agree that this PR deserves a split, so I will do it in the coming days. Before we continue discussions in separate threads, here are some responses to your feedback:

>> [feat] Search bar no longer covers the actions menu (the menu one sets STDOUT / STDERR, download logs, etc.) button and RAM. Its default position used to sit right on top of both items, preventing user to see RAM or open the actions menu. It now opens below the top bar and, from that default position, animates aside when the actions menu opens. It is also draggable, with press-and-hold on the icon to reset its position.
> 
> This was always draggle. And now it blocks the progress status instead. I am not sure which is better. But that's why I made it draggable.

It's still draggable, and I apologize for not taking the progress status widget into account (Maybe my use cases were not complex enough to make the widget visible). While we are at it, I feel the need to allow user to reset search bar location, which is why I introduced that Spotlight-like hold to reset feature. May I have some feedback on this as well?

>> [feat] CPU usage toggle. Brings back the older per-core (Linux style) display as an opt-in, alongside the current whole-CPU % default. Stored per profile.
>
> This needs to also take affect on the alerts. It should be it's own PR. I am actually unsure if this would be valuable to add. No one has asked for it.

The re-introduction of top-style CPU usage came up because I relied on the old Dozzle to find out how many CPU cores did my service utilize , and this metric is especially useful when the specs of my fleet of machines are not equal -- I had machines with 16, 24, and 32 vCPUs, and yet this metric will likely remain in the same range because it's independent of how many CPU cores / vCPUs are available on a given rig. I understand that this is a niche use case, so I had it default to off.

>>[fix] Actions menu works on touch now. It was hover-only and auto-hid on tap-focus, so it never opened on mobile. It opens on tap now and only uses hover where the device supports it.
>
> We can put all the fixes related to mobile in one PR.

Sounds good! I will combine it with the small improvements and fixes you mentioned at the end. 

>> [feat] Optional fullscreen settings popup (default off, flagged experimental). The /settings route is unchanged, and  the popup reuses the same panel.
>
> Why? How often are people going to settings page that this needs to be triggered from any page? Could we discuss that? I'd want to know what needs to be changed often.

There are two reasons to introduce this popup-style for me:

- I do override color scheme (Dark Mode / Light Mode) from time to time, and I prefer tweaking the display settings against my current container. 
- It's a one way user flow right now. Currently, when the user wishes to change style (or any settings), then they must navigate away by clicking on Dozzle icon (which my instinct was finding a menu dropdown or settings button, we don't have it right now), then use the settings on the top right to get to the /settings page. When i am done playing with the settings, I would usually go back to the container I was at to keep working, and this is where the user flow becomes sub-optimal -- I have to manually find which container I was at on the sidebar or use browser's back button **twice** to go back. In case we introduce more customization options such as color palettes for logs and new time formats, I can see how user might be bothered by this. 
 
> I played around with the code. Some of it is good. I don't agree that the drop down should have any items not related to that specific container. That dropdown is about that page. Having things like reset width or show settings are completely irrelevant to everything else.

I agree. I will put a hold on this dropdown change. With that said, this is the same "I tried to reach for a settings menu" situation, so I put the features in the dropdown as an experiment. 

Thanks again for your feedback!

amir20:
>I do override color scheme (Dark Mode / Light Mode) from time to time, and I prefer tweaking the display settings against my current container. 

I think it would be super cool for the keyboard shortcut `command + k` to allow for power user toggles like theme.  Then people can just do `cmd + k` and `/theme dark` or just `switch theme: dark`. Similar to vscode. I have been wanting to add this for a while. It would be great for power users who want to make quick changes. 

If you do this, then you no longer need to even worry about settings be a popup. Adding experimental features isn't something I do often.

LBYPatrick:
Good idea! I will get to it after I file the small features PR. 

I am thinking about implementing this on top of the 'search containers' popup to make it dual purpose just like VSCode. `Cmd + K` brings up the find containers one while `Cmd + Shift + P` brings up the power mode, which differentiates by a leading `/` (Just like VSCode's `>`, but i don't like that). The end product will feel similar to [Raycast](https://www.raycast.com/).

Some quick command candidates in my mind:
- `/theme <dark | light | auto>`
- `/compact <true | false>`
- `/soft-wrap <true | false>`
- `/fontsize <int>`
- `/reset` -> resets everything to default
- `/export` -> copies ALL preferences to clipboard as a json
- `/import <json | some URL with application JSON>` <- From that `/export`
- ...

amir20:
Yup. I was thinking the same thing. I wouldn't over complicate though. Discovery of these features is challenging. Perhaps at first, you can just search for containers or commands. All the same shortcut. 

Either way I think this would a nifty feature. Just remember most people are not power users.

LBYPatrick:
Sure, I'll keep the initial design straightforward.

amir20:
I was curious on how this would look. I created a PR https://github.com/amir20/dozzle/pull/4861. I wasn't planning on merging but maybe it's a good start for you.

PR #4861, opened 2026-07-23
feat: Command palette in Cmd+K search

amir20:
Extends the existing Cmd+K global search into a VS Code-style command palette. No new shortcut (Shift+Meta+K is taken by browsers). Commands show alongside containers/services/stacks with unified arrow/enter navigation.

## Commands
- **Container actions** (only on a container page, state-gated): Restart / Stop / Start / Update
- **Settings toggles**: light/dark theme, compact, timestamps, soft wrap, stopped containers, stdout/stderr labels, smaller scrollbars
- **Navigation**: Open settings

On an empty query the container-context actions surface immediately, so Restart is one keystroke away on a container page.

## Implementation
- New `assets/composable/commands.ts` — `useCommands()` registry, context-aware and reactive to route/settings
- `FuzzySearchModal.vue` — added a Commands section above Containers; `selectedIndex` now spans a single flat list so keyboard nav works across both. Enter runs `perform()`; Alt+Enter still pins container columns
- `locales/en.yml` — new `command-palette` block (other locales fall back to en)

## Tests
- 2 new modal tests (command match by keyword, running a command toggles its setting)
- typecheck clean, 120 frontend tests pass


assets/composable/commands.ts
```diff
@@ -0,0 +1,180 @@
...
+  const { start, stop, restart, update } = useContainerActions(currentContainer as Ref<Container>);
+
+  const commands = computed<Command[]>(() => {
+    const list: Command[] = [];
+
+    const container = currentContainer.value;
+    if (container) {
+      const name = container.name;
+      list.push({
+        id: "container.restart",
```
amir20, line 71: should this only be exposed if restart is supported? Actions are disabled by default

LBYPatrick:
Cool! I will play with it on my machine now. Will get back to you soon.

LBYPatrick:
Hi Amir, this is a very cool addition and I got addicted to it almost immediately 😮 

For the commands, in addition to Fuzzy Match, I am more familiar with the `/` prefix for explicit command requests, like this: 

https://github.com/user-attachments/assets/7fc874e5-9c4c-4dd9-a2b2-74e15cd205b1

This change currently lives in my fork under the same 'command-palette' branch, which you can check out [here](https://github.com/LBYPatrick/dozzle/tree/command-palette) 

And for the sidebar access, once I file the PR that introduces Cmd + K on the sidebar, this feature will have app-wide access 😄

amir20:
> And for the sidebar access, once I file the PR that introduces Cmd + K on the sidebar, this feature will have app-wide access 😄

Not sure I understand. It already is available app wide. No need to do anything. The keyboard shortcut should work on any page.

amir20:
Also, I am not particularly opinionated about `/`. `/foo` makes sense for chat style tools where there is a clear difference between a message and action. The search box does fuzzy matching so it doesn't really need an extra command. But it is nice to just show all the actions at the same time.

amir20:
Did a lot of testing. Going to merge this so I can stop dealing with conflicts.

PR #4862, opened 2026-07-24
feat(ui): small improvements and polishments

LBYPatrick:
Per discussion on #4856, here is the smaller PR consisting of smaller tweaks:

- fix: Clipboard features were not available under non-secure contexts
- fix:  document scrollbar gutter for the container logs was not always kept, causing a very subtle layout flick when the log finishes loading
- fix: Log level chips under light mode had black text on dark color chips, which violates WCAG's requirement on minimum contrast. With the exception of `warn`, other chips now use white text under light mode to be compliant.
- fix: The Command + K popup had a clipped shadow at the bottom edge. 
- fix: On mobile, container status bar was above the container list pane, which should be below.
- fix: Sidebar expanded to a non-consistent width, sometimes clipping Dozzle logo
- feat: image tags are copy-able by click

For mobile and sidebar, the new features are held back while the fixes are kept.


amir20:
Reviewed this with claude and tested it. looks good.

amir20:
After releasing and testing on mobile I noticed the stats is hidden sometimes. Not always but this wasn't like that before. So I'll have to revert. https://github.com/amir20/dozzle/pull/4867/changes

PR #4872, opened 2026-07-28
ci: attach an SBOM attestation to the published image

kobihikri:
Hi, and thanks for Dozzle.

`.github/workflows/deploy.yml` publishes the image, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

Dozzle is given the Docker socket — that is how it reads logs — which makes it one of the most privileged containers on a typical host. An image handed the socket is a good candidate for being verifiable first.

The change is two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth stating: `mode=max` records the full build including build arguments, so `provenance: true` is the smaller option if any have ever been sensitive; and attestations add an extra manifest to the index, which the registry supports.

No SLSA level claimed — the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.


amir20:
Thanks for the writeup. A few things before I take this:

BuildKit provenance isn't signed, so it doesn't really answer "was this built by this workflow". Anyone can attach a provenance blob claiming the same repo and tag. If the goal is verification, that needs `actions/attest-build-provenance` (sigstore, `id-token: write`) or cosign. Also build-push-action already defaults to min-mode provenance, so `provenance: true` is close to a no-op here. The SBOM is the part that actually adds something.

My concern is the extra `unknown/unknown` manifests in the index. Dozzle ships 4 platforms including arm/v6 and a lot of users are on older Docker or containerd, or pulling through registry mirrors. That's where attestation manifests tend to cause confusing failures, and it lands on me in issues.

Happy to merge if you drop it to `provenance: mode=min` and keep `sbom: true`. Also restore the trailing newline at the end of the file.

If you want to do the signed attestation properly instead, open that as a separate PR and I'll look.

kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

amir20:
I think this is right. I had to do research too. So sounds like `provenance: mode=max ` can be dropped. Title needs to be updated and keep SBOM. 

Is that right?  If so let's make those changes and I can merge.

kobihikri:
Yes, that's right. Done — the diff is now a single line:

```diff
           push: true
+          sbom: true
           context: .
```

`+1/-0`, title updated to match (it's an SBOM change now, not a provenance one), trailing newline already restored in the earlier commit.

**On your `unknown/unknown` concern — I checked, and I think it's already moot, which I should have verified before rather than after.**

Because provenance is on by default for public repos, those attestation manifests are *already* in your published index. From `amir20/dozzle:latest` just now:

```
linux/amd64
linux/arm/v6
linux/arm/v7
linux/arm64
unknown/unknown   attestation-manifest
unknown/unknown   attestation-manifest
unknown/unknown   attestation-manifest
unknown/unknown   attestation-manifest
```

Four platforms, four attestation manifests, today, without this PR. `sbom: true` adds SBOM data *into* those existing attestation manifests rather than adding new entries to the index — so the index shape your older-Docker and mirror users see doesn't change.

That means the compatibility risk you were weighing is one you're already carrying, and this PR doesn't increase it. If you'd rather *reduce* it, the lever is `provenance: false`, which would remove those four entries — happy to open that as a separate PR if it's useful, though it's a different decision from this one and I wouldn't fold it in here.

**And you were right on the substance twice over.** Provenance being unsigned means it doesn't answer "was this built by this workflow" — `actions/attest-build-provenance` or cosign is the real answer there, and I shouldn't have implied otherwise. I've left that out entirely; if you ever want it, it's a separate change with `id-token: write`.

Thanks for taking the time to research it rather than just closing this — the correction improved the PR.

*Disclosure: AI-assisted (Claude Opus 5). I verified the manifest index, the Docker attestations docs and the diff myself before posting.*

Issue #4858, opened 2026-07-23
Add support for a configurable JWT signing secret


OGIRAUDOUX-77:
### Describe the feature you would like to see

I would like Dozzle to support a configurable JWT signing secret through an environment variable (for example `DOZZLE_JWT_SECRET`) or a secret file (for example `DOZZLE_JWT_SECRET_FILE`).

Currently, Dozzle generates a new signing key every time it starts. As a result, all previously issued JWTs become invalid after a restart. This makes long-lived integrations difficult, as every restart requires generating a new token and updating the client configuration.

Having an optional persistent signing secret would allow JWTs to remain valid across restarts while preserving the current behavior for users who do not configure it.

### Describe how you would like to see this feature implemented

The implementation could be fully backward compatible:
- If `DOZZLE_JWT_SECRET` (or a similar variable) is provided, use it as the JWT signing secret.
- If `DOZZLE_JWT_SECRET_FILE` is provided, read the secret from the specified file (useful for Docker Secrets and Kubernetes Secrets).
- If neither option is configured, continue generating a random signing key at startup, preserving the current behavior.

This would allow users to opt into persistent JWTs only when needed.

### Describe any alternatives you've considered

The current workaround is to manually generate a new JWT after every Dozzle restart and update the client configuration.

In my case, I am using the Dozzle MCP server with Open WebUI. Every time Dozzle restarts, the stored JWT becomes invalid, causing the MCP integration to stop working until I manually generate and configure a new token.

Automatically re-authenticating is possible for some clients, but not all integrations support it. A configurable signing secret would provide a simple and reliable solution for long-lived API and MCP integrations.

amir20:
The signing key isn't random and doesn't change on restart. It's derived deterministically from the users in your `users.yml`:

```go
// internal/auth/simple.go
h := sha256.New()
for _, user := range userDatabase.Users {
    h.Write([]byte(user.Password))
    h.Write([]byte(user.RolesConfigured))
}
tokenAuth := jwtauth.New("HS256", h.Sum(nil), nil)
```

Restarting Dozzle produces the same key, so an existing JWT stays valid. With the default `--auth-ttl=session` there's no expiry claim either, so tokens don't time out.

What does invalidate a token is a change to the hash inputs: a password gets re-hashed or changed, roles change, or a user is added/removed. If your token stopped working after a restart, something in `users.yml` changed between the two runs. Can you check whether the password hash for that user is identical before and after?

[...]

PR #4886, opened 2026-08-04
fix: derive jwt signing key from users in a stable order

TowyTowy:
In #4858 the answer was that the signing key is deterministic and a JWT survives a restart unless something in `users.yml` changes. That holds for a single user, but not for more than one.

`NewSimpleAuth` builds the sha256 by ranging over `userDatabase.Users`, which is a map, and Go randomizes map iteration order. With two or more users the digest comes out different on most starts, so the key rolls and every session cookie and API/MCP token dies on restart. Feeding the same 4-user map through the current loop 200 times produced 3 distinct keys.

Sorting by username keeps the exact same hash inputs and makes the result reproducible. A single-user `users.yml` derives the same key it did before, so those installs see no change; multi-user installs get logged out once on upgrade and then stay logged in.

Tests cover both directions: a token issued before a restart still verifies afterwards, and changing a role still invalidates it.

Prepared with AI assistance (Claude); fail-before/pass-after and `go test -race` verified before opening.

### DIFF
--- a/internal/auth/simple.go
+++ b/internal/auth/simple.go
@@ -3,7 +3,9 @@ package auth
 import (
 	"crypto/sha256"
 	"errors"
+	"maps"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/go-chi/jwtauth/v5"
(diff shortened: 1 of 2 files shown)


amir20:
Hmm I would've expected the order to never change but I guess Go could do whatever.

PR #4893, opened 2026-08-06
fix: disambiguate containers with duplicate names across groups

lolen:
When two groups (dev.dozzle.group) contain containers with the same name, the main container table and the container title dropdown show them as identical, indistinguishable rows.
  
- ContainerTable.vue: show the group name under the container name, and use it as a secondary sort key
- ContainerTitle.vue: scope the "switch to" dropdown to containers within the same group instead of matching by name alone across all groups; show the host label when switching between instances on different hosts


ready4sea:
Great feature. Would have been nice to also add the group name as an option in webhooks for notifications (https://dozzle.dev/guide/alerts-and-webhooks).

PR #4898, opened 2026-08-08
ci: extend commit SHA pinning to every workflow action

Totara-thib:
You already pin some of these, `docker/setup-buildx-action` is on a SHA with a version comment and `docker/login-action` sits on an exact tag. This PR finishes the job: every action in the seven workflows is now pinned by full commit SHA, one commit, 53 lines, versions stay exactly where they were.

The reason to close the gap: a floating tag is a movable pointer. Whoever controls an action repository can point the tag at new code, and the next `deploy` run executes it with the Docker Hub credentials and the TTL certificate secrets in scope. The tj-actions/changed-files incident (CVE-2025-30066) did exactly that, existing tags rewritten on a popular action to leak CI secrets. A SHA cannot be retargeted after the fact.

Each pin keeps the version as a trailing comment so it can be checked against the action's releases page. Renovate resolves SHA pins and keeps bumping them as digests, your weekly non-major automerge flow is unaffected.

One heads-up: if you ever restrict allowed actions in the repo settings with tag patterns like `owner/action@v1`, those patterns stop matching SHA refs and workflows fail at startup, the fix is `owner/action@*` there. Nothing to do if no such restriction is configured.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -27,13 +27,13 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
 
       - name: Run Claude Code Review
         id: claude-review
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@6b082c41935b4c8a3b8b0ef85ba4ba4d9eeb8975 # v1.0.189
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           use_sticky_comment: true
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -30,13 +30,13 @@ jobs:
       actions: read # Required for Claude to read CI results on PRs
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@v1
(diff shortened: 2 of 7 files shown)


amir20:
I am afraid this wouldn't work. I use renovate to automatically update all my deps and it would just convert all the SHA pins to tag again. I am not sure if Renovate supports this.

amir20:
one correction on "your weekly non-major automerge flow is unaffected". it does get affected.

renovate classifies a SHA bump as a `digest` update, and both presets in `.github/renovate.json` match only `["minor", "patch"]`:

- `group:allNonMajor`
- `:automergeMinor`

so digest updates fall through both. once every action is a SHA, that is one manual PR per action per week instead of the one grouped automerge we get today.

opened #4899 to add the packageRule for it. merging that first, then this.

```json
{
  "matchManagers": ["github-actions"],
  "matchUpdateTypes": ["digest", "pin"],
  "groupName": "github actions",
  "automerge": true
}
```

rest of the reasoning is sound, tj-actions is a fair thing to point at. no allowed-actions restriction configured here so that heads-up is moot.

Totara-thib:
> one correction on "your weekly non-major automerge flow is unaffected". it does get affected.
> 
> renovate classifies a SHA bump as a `digest` update, and both presets in `.github/renovate.json` match only `["minor", "patch"]`:
> 
> * `group:allNonMajor`
> * `:automergeMinor`
> 
> so digest updates fall through both. once every action is a SHA, that is one manual PR per action per week instead of the one grouped automerge we get today.
> 
> opened #4899 to add the packageRule for it. merging that first, then this.
> 
> ```json
> {
>   "matchManagers": ["github-actions"],
>   "matchUpdateTypes": ["digest", "pin"],
>   "groupName": "github actions",
>   "automerge": true
> }
> ```
> 
> rest of the reasoning is sound, tj-actions is a fair thing to point at. no allowed-actions restriction configured here so that heads-up is moot.

You are right and I overstated that. Digest updates do fall through `group:allNonMajor` and `:automergeMinor`, I should have read the presets instead of assuming. The packageRule in #4899 is the clean fix, grouped and automerged like before. Thanks for correcting it properly and for merging.

PR #4900, opened 2026-08-08
feat: publish optional alpine image variant

### LINKED ISSUE #4897: Optional shell-bearing image variant (e.g. `:alpine`) for platforms that wrap the entrypoint

An optional published tag — e.g. `amir20/dozzle:alpine` — that is the same Dozzle binary on a minimal base with a shell, alongside the existing scratch-based default.

To be clear up front: **this is a convenience request, not a bug, and there is a working workaround.** I wouldn't reopen #4576 over it. The scratch image is the right default and I'd like it to stay the default.

### Why

Several self-hosting platforms integrate optional features by bind-mounting a `#!/bin/sh` wrapper over the container's entrypoint and re-exec'ing the real binary through `$ORG_ENTRYPOINT`. Unraid's built-in per-container Tailscale toggle is the one I hit; sidecar/init injectors in other orchestrators use the same trick. A `FROM scratch` image has no interpreter, so the container dies with:

```
exec /opt/unraid/tailscale: no such file or directory
```

which reads as a missing hook rather than a missing shell, and sends people down the wrong path — see #4576, where the reporter (reasonably) concluded Dozzle was incompatible with Unraid Tailscale.

I've posted the full root cause and a no-image-change workaround on #4576: bind-mount a static busybox as `/bin/sh` and a `/usr/bin` containing static `jq` + the `tailscale` binaries, and the wrapper runs fine against stock `amir20/dozzle:latest`. That works and I'm running it in production, so nothing is blocked.

The ask is only that a `:alpine` tag would turn a half-day of investigation into changing one character in a template, for anyone else who hits this.

A second stage in the existing build publishing the same binary on `alpine:latest`, tagged `:alpine` (and `:vX.Y.Z-alpine`), with the default `:latest` unchanged. Roughly:

```dockerfile
FROM alpine:latest AS alpine-variant
COPY --from=<existing build stage> /dozzle /dozzle
ENTRYPOINT ["/dozzle"]
```

Docs could note it's for platforms that need a shell in the image, and that `:latest` remains the recommended one.

- **The bind-mount shim above.** What I'm actually using. Works, but every user has to rediscover it, and it needs the static-binary detail to be right (`ldd` misleadingly reports static-pie binaries as fine, so `readelf -l` / `readelf -d` is the real check).
- **Repackaging the image locally onto Alpine.** I did this first. It works but forks the image — Dozzle releases stop arriving through the platform's auto-update and every upgrade becomes a manual rebuild. I abandoned it for exactly that reason, which is what motivated the shim.
- **Doing nothing.** Entirely reasonable. If you'd rather not carry a second tag, closing this with a docs line pointing at the #4576 workaround would still help the next person.

amir20:
Closes #4897

The default image is `FROM scratch`, so there is no shell in it. Some platforms add optional features by bind-mounting a `#!/bin/sh` wrapper over the container entrypoint and re-exec'ing the original one through `$ORG_ENTRYPOINT`. Unraid's per-container Tailscale toggle does this, and so do a few sidecar/init injectors. With no interpreter in the image the exec fails and the error names the wrapper rather than the missing shell:

```
exec /opt/unraid/tailscale: no such file or directory
```

which is what sent #4576 down the wrong path.

This publishes the same binary on an Alpine base as `:alpine` and `:vX.Y.Z-alpine`. `:latest` stays scratch.

## Changes

- `Dockerfile`: new `alpine` stage placed above the scratch stage, so the last stage remains the default build target and untargeted builds (`make docker`, compose, int tests) are unaffected
- `deploy.yml`: second metadata + build step pushing `alpine`, `vX.Y.Z-alpine`, `vX.Y-alpine`, `vX-alpine` on the same four platforms
- `dev.yml`: same for branch/PR images, so this is testable as `amir20/dozzle:pr-XXXX-alpine`
- FAQ entry describing the symptom and pointing at the variant

The binary is `CGO_ENABLED=0`, so musl vs glibc is not a factor. `ca-certificates.crt` and `/data` are copied in since alpine ships neither.

## Testing

```
$ docker build --target alpine -t dozzle-alpine .
$ docker run --rm dozzle-alpine --version
dev

$ docker run --rm --entrypoint /bin/sh dozzle-alpine -c 'echo shell-ok'
shell-ok

# entrypoint wrapper, same shape as Unraid's
$ docker run --rm -e ORG_ENTRYPOINT=/dozzle -v ./wrapper.sh:/wrap.sh \
    --entrypoint /wrap.sh dozzle-alpine --version
wrapper ran
dev
```

Default target still builds scratch and still has no shell:

```
$ docker run --rm --entrypoint /bin/sh dozzle-default -c 'echo x'
exec: "/bin/sh": stat /bin/sh: no such file or directory
```

Size: 64.4MB scratch, 73.1MB alpine.

## Tradeoff

Scanners will now report Alpine CVEs against the variant, and patching them means cutting a release or accepting a stale base between releases. That is the ongoing cost of carrying it. An alternative that avoids a second tag entirely is copying a static busybox into the existing scratch image (~1MB), which fixes this for everyone without a variant, at the cost of a shell in the default image. Happy to switch to that if you prefer it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -125,6 +125,18 @@ jobs:
             type=semver,pattern={{version}},prefix=v
             type=semver,pattern={{major}}.{{minor}},prefix=v
             type=semver,pattern={{major}},prefix=v
+      - name: Docker meta for alpine variant
+        id: meta-alpine
+        uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6.2.0
+        with:
+          images: |
+            amir20/dozzle
+            ghcr.io/amir20/dozzle
+          tags: |
+            type=raw,value=alpine
+            type=semver,pattern={{version}},prefix=v,suffix=-alpine
+            type=semver,pattern={{major}}.{{minor}},prefix=v,suffix=-alpine
+            type=semver,pattern={{major}},prefix=v,suffix=-alpine
       - name: Writing certs to file
         run: |
           echo "${{ secrets.TTL_KEY }}" > shared_key.pem
@@ -139,6 +151,19 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           build-args: TAG=${{ steps.meta.outputs.version }}
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Build and push alpine variant
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          push: true
+          sbom: true
+          context: .
+          target: alpine
+          platforms: linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64/v8
+          tags: ${{ steps.meta-alpine.outputs.tags }}
+          build-args: TAG=${{ steps.meta.outputs.version }}
+          labels: ${{ steps.meta-alpine.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
   git-release:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -164,8 +164,10 @@ jobs:
           tags: ${{ steps.meta-alpine.outputs.tags }}
           build-args: TAG=${{ steps.meta.outputs.version }}
           labels: ${{ steps.meta-alpine.outputs.labels }}
+          # No cache-to. This shares every stage up to the builder with the step
+          # above, which already exported them, and the alpine stage itself is
+          # three COPYs. Exporting again would only overwrite that step's index.
           cache-from: type=gha
-          cache-to: type=gha,mode=max
   git-release:
     needs: [buildx]
     name: Github Release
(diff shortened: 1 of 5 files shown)


amir20:
Pinned to `alpine:3.24` in 2448700b. Worth noting the other bases here aren't digest pinned either (`node:25.9.0-alpine`, `golang:1.26-alpine`), and `helpers:pinGitHubActionDigests` only covers actions. A minor pin matches the existing style and renovate's docker manager will bump it.

Dropped `cache-to` from the alpine step rather than splitting scopes. Separate scopes would make that step export the entire node+builder graph again into its own scope, roughly doubling GHA cache storage against the 10 GB repo limit. Sharing one scope is exactly what kept the export cheap (3 layers, 0.5s to prepare in run 1). Since the step shares every stage up to the builder with the step before it, and the alpine stage is three COPYs, there's nothing worth exporting.

Checked for the scratch regression you predicted and didn't find it. Run 1 step 1 was 3m07s, run 2 (after both exports had landed) was 3m23s. The 135s inside that is the go compile, which misses on every commit regardless because `TAG` embeds the short SHA and ends up in the ldflags. The alpine step ran 13.6s and 19s with `go build` and `pnpm build` both CACHED off local buildkit state.

Skipping `apk upgrade`. The alpine stage installs no packages, so it would only bump base busybox/musl, and a minor pin means `3.24` already resolves to the newest patch on every rebuild. Not worth an unpinned package set in the image.

PR #4907, opened 2026-08-13
fix: don't log an error when docker events listener exits cleanly

### LINKED ISSUE #4906: Idle stats-collector stop logs 'unexpected error while listening to docker events' with no error field

- [x] Completed (closest: #4622 has the same message but with `error=EOF` on a dead remote host; #4472 notes that `stopped container stats collector` after idle is expected)

### How is Dozzle deployed?

Standalone Deployment (local `docker.sock`, no agents / remote hosts / Swarm / Kubernetes)

### 📦 Dozzle version

v10.6.5 (`amir20/dozzle:latest` at the time)

```yaml
services:
  dozzle:
    image: amir20/dozzle:v10.6.5
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
    ports:
      - "8080:8080"
    environment:
      DOZZLE_NO_ANALYTICS: "true"
```

Production also uses `DOZZLE_AUTH_PROVIDER=simple`; auth does not appear related.

After the last browser tab disconnects, the stats collector correctly stops ~6 hours later (`timeToStop`). At that same instant Dozzle also logs:

```
level=error message="unexpected error while listening to docker events"
```

There is **no `error=` field**. Dozzle stays up, the UI still works, and a later login does not require a restart. This is not a Docker daemon disconnect (contrast #4622, which includes `error=EOF` and `docker store unexpectedly disconnected from docker events`).

**Steps**

1. Start the compose file above.
2. Open the UI until the container list and CPU/memory charts appear (this starts the stats collector and its `/events` subscription).
3. Close every Dozzle tab.
4. Wait ~6 hours (do not restart the container or Docker).
5. Inspect `docker logs`.

**Expected:** only the intentional idle stop, e.g. `stopped container stats collector` at info.

**Actual:** that info line is paired with `unexpected error while listening to docker events` at error, with no `error=` key.

**Why:** `ContainerEvents` returns `nil` on a normal cancel:

https://github.com/amir20/dozzle/blob/v10.6.5/internal/docker/client.go

```go
case <-ctx.Done():
    return nil
```

The stats collector then treats anything other than `context.Canceled` as unexpected:

https://github.com/amir20/dozzle/blob/v10.6.5/internal/docker/stats_collector.go

```go
err := sc.client.ContainerEvents(ctx, events)
if !errors.Is(err, context.Canceled) {
    log.Error().Str("host", sc.client.Host().Name).Err(err).Msg("unexpected error while listening to docker events")
}
```

`errors.Is(nil, context.Canceled)` is false, and zerolog `.Err(nil)` omits `error=`, which matches the logs.

Suggested fix: skip the error log when `err == nil` as well, e.g. `if err != nil && !errors.Is(err, context.Canceled)`.

This still appears on current `master` (`ContainerEvents` still returns `nil` on `ctx.Done()`; collector check is unchanged).

### 💻 Environment

Linux Docker host, Engine API via bind-mounted `/var/run/docker.sock`. Single local engine (`Connected to Docker` / `clients=1`). Hostname in logs is the daemon name from `docker info` (`fastapi-webapp`). Happy to paste `docker system info` if useful; this looks like a code-path issue rather than a daemon/config one.

### 📜 Dozzle logs

Pattern repeats after UI use: login, then ~6 hours later the pair. Example where the gap is exactly 6 hours:

```
level=info version=v10.6.5 user=admin time=2026-07-10T01:10:20Z message="Token created"
level=info version=v10.6.5 host=fastapi-webapp time=2026-07-10T07:10:41Z message="stopped container stats collector"
level=error version=v10.6.5 host=fastapi-webapp time=2026-07-10T07:10:41Z message="unexpected error while listening to docker events"
```

Same pair ~6.3h after another login (no `error=` field):

```
level=info version=v10.6.5 user=admin time=2026-07-08T04:56:33Z message="Token created"
level=info version=v10.6.5 host=fastapi-webapp time=2026-07-08T11:15:21Z message="stopped container stats collector"
level=error version=v10.6.5 host=fastapi-webapp time=2026-07-08T11:15:21Z message="unexpected error while listening to docker events"
```

Dozzle did not exit. Later logins still succeed (`Token created`) without a container restart. The only process stop in a multi-week log was an unrelated graceful shutdown on deploy.

Gaps longer than 6 hours match a tab left open (SSE keeps the collector alive until the last disconnect, then the timer starts).

amir20:
`ContainerEvents` returns `nil` on a normal `ctx.Done()`, but callers only guarded with `errors.Is(err, context.Canceled)`. `errors.Is(nil, context.Canceled)` is false, so a clean shutdown logged an error with no `error=` field.

Fixed in both call sites:
- `internal/docker/stats_collector.go:121` (idle stats collector stop, the one in the issue)
- `internal/container/container_store.go:104` (same pattern)

Fixes #4906

### DIFF
--- a/internal/container/container_store.go
+++ b/internal/container/container_store.go
@@ -101,7 +101,7 @@ func (s *ContainerStore) checkConnectivity() error {
 		go func() {
 			log.Debug().Str("host", s.client.Host().Name).Msg("docker store subscribing docker events")
 			err := s.client.ContainerEvents(s.ctx, s.events)
-			if !errors.Is(err, context.Canceled) {
+			if err != nil && !errors.Is(err, context.Canceled) {
 				log.Error().Err(err).Str("host", s.client.Host().Name).Msg("docker store unexpectedly disconnected from docker events")
 			}
 			s.connected.Store(false)
--- a/internal/docker/stats_collector.go
+++ b/internal/docker/stats_collector.go
@@ -118,7 +118,7 @@ func (sc *DockerStatsCollector) Start(parentCtx context.Context) bool {
 		defer close(events)
 		log.Debug().Str("host", sc.client.Host().Name).Msg("starting to listen to docker events")
 		err := sc.client.ContainerEvents(ctx, events)
-		if !errors.Is(err, context.Canceled) {
+		if err != nil && !errors.Is(err, context.Canceled) {
 			log.Error().Str("host", sc.client.Host().Name).Err(err).Msg("unexpected error while listening to docker events")
 		}
 		sc.forceStop()


chriscarrollsmith:
Looks right to me!

PR #4911, opened 2026-08-14
feat(cloud): stream windowed container metrics to Dozzle Cloud

amir20:
Ships per-container resource metrics to Dozzle Cloud over the gRPC connection that already carries log lines.

Cloud today stores log lines and *alerts*, but no resource time series — there is no way to answer "was this container pegged at 100% CPU for the ten minutes before it OOMed?". This adds the producer half. The consumer half (storage in VictoriaMetrics) is a companion PR in the Cloud repo, which is blocked on this one merging so it can bump the module.

## How it works

Raw stats (~1Hz from Docker) are folded into 30-second windows and pushed as an unsolicited `StatsBatch` on the existing `ToolStream`, alongside `LogBatch`.

**One goroutine for the whole instance.** `logStreamer` needs a goroutine per container because each holds an open log stream; stats already arrive multiplexed on one channel, so there is nothing to parallelise. A 50-container host sends one ~4KB message every 30s instead of 1500 raw samples.

**Gauges carry mean *and* peak.** A 30s average flattens exactly the spikes an incident chart is read for, so `cpu_percent_max` / `memory_percent_max` ride along — cheap, and the interesting half of the signal.

**Counters send their last value, not a delta.** Network/disk totals are cumulative since container start. Sending the raw counter lets PromQL `rate()`/`increase()` handle the reset that happens when a container is recreated, which a pre-computed delta could not.

**CPU is normalised before it leaves the host.** `Stat.CPUPercent` is per-core (100% == one full core). It is divided by `CPULimit` -> `Host.NCPU` -> 1, matching `notification/processing.go` — shipping it raw would disagree with both the UI and the metric-alert path. `cpu_cores` rides along so the raw figure stays recoverable.

**Series identity is `(host_id, container_name)`, with no `container_id`.** Container ids churn on every redeploy and each new value would strand its predecessor's series forever, breaking a CPU chart into disconnected segments. `ContainerStat` carries no host at all, so `localCloudHostService.SubscribeStats` stamps it at fan-in — otherwise two same-named containers on different hosts would collide.

## Privacy

The settings toggle now gates **both** logs and metrics, and its copy was rewritten to say so — "Stream container logs" understated what leaves the instance once metrics ride along. The card now spells out:

- what is sent while on: log lines as written, plus CPU/memory/network/disk averaged over 30s
- what is never sent: environment variables (verified — `env` is `reserved` in `cloud.proto` and nothing in `internal/cloud` reads `.Env`)
- what still works while off: configured alerts still deliver and remote container access still works; the dispatcher doesn't consult this toggle
- the per-container `dev.dozzle.cloud.min_level=disabled` opt-out

These use **new i18n keys** rather than edits to the existing ones. Editing `stream-logs`/`stream-logs-help` in place would leave the 15 untranslated locales asserting that only logs are shared — a stale privacy claim is worse than an English fallback.

Metrics reuse the toggle rather than adding a second one: a user who opted out of shipping log *contents* has opted out of shipping resource usage too. The streamer is bound to the connection lifetime, so it tears down on disconnect and re-evaluates the toggle on reconnect.

## Trade-off worth knowing about

An always-on subscriber **pins `DockerStatsCollector` on**. It is otherwise reference-counted and idles out 6h after the last subscriber, so today `docker stats` streams only run while someone is watching the UI or a metric alert is configured. Continuous metrics inherently change that. On a host with hundreds of containers this is not free; the `streamLogs` toggle is the escape hatch.

k8s is not covered — `newLocalCloudHostService` returns the bare `HostService` there, which doesn't satisfy the capability interface, so the streamer is skipped with a debug log (the same graceful-skip the log streamer already does).

## Testing

`go build ./... && go vet ./... && go test ./...` all green.

10 new tests in `stats_streamer_test.go` cover the aggregation directly (driving `observe`/`flush` rather than waiting on the 30s ticker): mean/peak correctness, counters taking the last value, all three CPU-normalisation fallbacks, state resetting between windows, empty windows sending nothing, disabled containers being skipped, same-name-different-host separation, and a container that starts mid-window being held then emitted once metadata catches up (or dropped after a grace window rather than leaking).

The write path was also verified end to end against a real VictoriaMetrics in the companion PR.

Note: the pre-commit hook reformatted a pre-existing misaligned `const` block in `client.go` — that's the small unrelated hunk.


amir20:
Thanks — the backpressure finding was real and is fixed in 094acf0b.

**The stall was worse than "cloud delivery lags".** You traced it exactly: `flush()` -> `refreshMeta()` -> `ListAllContainers()` ran inline on the goroutine draining `samples`, and because `DockerStatsCollector`'s dispatch does a blocking `stats <- stat` per subscriber, a wedged cloud consumer would eventually starve the **live UI stats broadcast** on that host. Metrics shipping degrading the UI is not a trade worth making.

Fixed in two layers:

1. **Metadata refresh moved off-loop.** It runs on its own goroutine and hands the snapshot back over a channel; `flush()` now does no I/O at all and works from the last-good snapshot. At most one refresh is in flight, so a host slower than the window doesn't queue work. Ownership stays with `run()`'s goroutine, so there's still no lock.
2. **The fan-in forwarder sends non-blocking.** Even if the aggregator wedged for some unrelated reason, cloud can no longer push backpressure into the collector. Dropping a sample nudges a 30s mean; stalling the UI doesn't.

**On the test — worth flagging.** My first regression test passed *both* with and without the fix, because the default 30s window meant `flush()` never ran during the test. It was decorative. The committed version drives a short window so `flush()` actually executes, and I verified it by reintroducing the inline refresh:

```
--- FAIL: TestStatsStreamer_SlowHostDoesNotStallSampleDrain (5.02s)
    sample drain stalled behind a slow host — backpressure would reach
    the stats collector and starve the live UI stats broadcast
```

Passes with the fix, fails without. Also added coverage for the snapshot actually being installed by the run loop and for only-one-refresh-in-flight. All 11 stats tests pass under `-race` (an earlier draft of the install test had a genuine data race reading `ss.meta` from the test goroutine — that's why it now asserts on emitted batches instead of internal state).

---

**Also in this push: the settings copy.** The toggle gates metrics as well as logs now, but still said "Stream container logs", which understated what leaves the instance. Replaced with a privacy block that:

- names both data types explicitly (log lines as written; CPU/memory/network/disk averaged over 30s)
- states environment variables are never sent — verified, not assumed: `env` is `reserved` in `cloud.proto` and nothing in `internal/cloud` reads `.Env`
- explains what still works when it's off (configured alerts still deliver, remote container access still works — the dispatcher doesn't consult this toggle)
- surfaces the per-container `dev.dozzle.cloud.min_level=disabled` opt-out

These are **new i18n keys** rather than edits to `stream-logs`/`stream-logs-help`. Editing in place would have left the 15 untranslated locales asserting that only logs are shared — a stale privacy claim is worse than an English fallback.

`go test ./...`, `pnpm run typecheck`, `pnpm run test` (121), and `pnpm run build` all pass.

PR #4915, opened 2026-08-16
feat(logs): merge Dozzle Cloud alerts into the log stream

amir20:
Scrolling back through a container now surfaces the alerts that fired in that
window, anchored to the line that tripped them.

```
3:48 PM  Ut ullamcorper, ligula eu tempor congue, eros est euismod turpis
3:49 PM │ ALERT  Container restarting on a failed health check  35 events   [Details ›] [View in Dozzle Cloud]
3:49 PM  Maecenas fermentum consequat mi
```

> [!NOTE]
> Stacked on #4914 (which is stacked on #4913). Review those first — this PR's
> base is #4914's branch, so the diff here is UI only.

### Anchoring is by log id, not timestamp

Cloud stores the FNV-32a hash of the line that triggered each alert — the same
id `LogEntry.id` carries and that "Copy permalink" uses. `mergeAlerts` splices
the alert directly after that line.

Sorting by time cannot do this, and that's not a detail: an alert shares a
millisecond with the line that caused it, so ordering by timestamp would place
it on either side depending on sort stability. Which line tripped the rule is
the entire point of the feature. `lands after the trigger line even when
timestamps are identical` is the test for it.

Falls back to timestamp position for metric and event alerts (no trigger line)
and for an alert whose trigger sits outside the loaded range — checked, so an
alert never silently vanishes.

### Dedupe is keyed on (alertId, anchor)

Cloud folds follow-up batches into the *original* alert row, so one incident
legitimately marks every window it stayed active in. Keying the seen-set on
`alertId` alone would mean scrolling newest→oldest loads a follow-up anchor
first and then swallows the origin as a duplicate — hiding exactly the thing
the user scrolled back to find. There's a test named for this.

Origin anchors render the row below; follow-ups render one quiet "still firing"
line, so a long incident doesn't bury the logs.

The set is cleared when `params` or `containers` change, since the viewer clears
its messages on a stream switch and the set would otherwise outlive what it was
tracking.

### The row itself

Three commits here, because the first two attempts were wrong when seen against
real data:

1. The icons never rendered — written as `class="iconify lucide--*"`, but this
   project resolves icons as components via `unplugin-icons` and has no lucide
   collection installed. Silently inert.
2. `@apply bg-error/10` compiles to a pair — a solid `background-color`
   fallback plus the intended color-mix tint — and through the scoped-style
   transform the solid one reached the page. The 10% tint rendered at 100%,
   and `text-error` on top of it made the text invisible.
3. The tinted band was still too loud for what it carried. **Final treatment is
   one line**: a red chip and rail against the untouched log background, with
   containers / held-back / investigation behind a `Details` toggle. The toggle
   is omitted when there's nothing to expand — a control that opens an empty
   panel is worse than no control. Actions are right-aligned so the headline
   starts at the same x-position on every alert, which is what makes a column
   of them scannable.

`alerts.level` is free text and defaults to `''`, so unknown levels read as
error rather than falling through to a neutral grey that looked like an ordinary
log line.

### Failure behaviour

`fetchAlerts` never rejects, and `withAlerts` still guards. This runs after the
log lines have been fetched — a cloud outage must cost the user their alerts,
never their logs.

### Verification

- `pnpm run typecheck` and `pnpm run build` pass
- 14 tests on the merge (anchoring, same-millisecond ordering, cross-container
  id collisions, both fallback paths, dedupe in both directions, degrade-on-
  failure) and 12 on the row (one-line layout, Details toggle open/close/absent,
  level defaulting, conditional affordances)
- **Pre-existing failure, not from this stack:** 3 snapshot tests in
  `EventSource.spec.ts` fail on `master` too. Left alone.
- **Not verified:** the rendered result. I can't see the running UI, so
  proportions inside the real row striping and container-name column are
  unconfirmed — particularly whether the right-aligned buttons crowd the
  headline on a narrow pane.

### Stack

- #4913 — proto + codegen
- amir20/doligence#338 — the `GetAlerts` server
- #4914 — cloud client + `/api/cloud/alerts`
- **this PR** — the stream merge
- #4916 — incident duration on the summary
- #4917 — scroll-gutter markers

### DIFF
--- a/assets/auto-imports.d.ts
+++ b/assets/auto-imports.d.ts
@@ -62,6 +62,7 @@ declare global {
   const effectScope: typeof import('vue').effectScope
   const escapeHtml: typeof import('./utils/index').escapeHtml
   const extendRef: typeof import('@vueuse/core').extendRef
+  const fetchAlerts: typeof import('./composable/cloudAlerts').fetchAlerts
   const flattenJSON: typeof import('./utils/index').flattenJSON
   const flattenJSONToMap: typeof import('./utils/index').flattenJSONToMap
   const formatBytes: typeof import('./utils/index').formatBytes
@@ -102,6 +103,7 @@ declare global {
   const mapWritableState: typeof import('pinia').mapWritableState
   const markRaw: typeof import('vue').markRaw
   const menuWidth: typeof import('./stores/settings').menuWidth
+  const mergeAlerts: typeof import('./composable/cloudAlerts').mergeAlerts
   const nextTick: typeof import('vue').nextTick
   const onActivated: typeof import('vue').onActivated
   const onBeforeMount: typeof import('vue').onBeforeMount
@@ -218,6 +220,7 @@ declare global {
   const useClipboard: typeof import('@vueuse/core').useClipboard
   const useClipboardItems: typeof import('@vueuse/core').useClipboardItems
   const useCloned: typeof import('@vueuse/core').useCloned
+  const useCloudAlerts: typeof import('./composable/cloudAlerts').useCloudAlerts
   const useCloudConfig: typeof import('./composable/cloudConfig').useCloudConfig
   const useCloudLogSearch: typeof import('./composable/cloudLogSearch').useCloudLogSearch
   const useColorMode: typeof import('@vueuse/core').useColorMode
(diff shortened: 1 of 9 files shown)


amir20:
Pushed a fix after testing against real data — the row was reading as one more log line with slightly different text.

Two causes, both invisible to the unit tests:

1. **The icons never rendered.** They were written as `class="iconify lucide--*"`, but this project resolves icons as components via `unplugin-icons` and has no lucide collection installed. The markup was silently inert — no error, just nothing.
2. **The border was grey almost every time.** `alerts.level` is free text and defaults to `''` for any alert that never went through a summarizer, and `''` fell through to the neutral case. So the colour meant to distinguish an alert was the one thing that never showed up.

The row is now a tinted band with a heavy left rule and an `ALERT` label rather than a bare left border, and unknown/empty levels read as error. Added three tests for the level defaulting, since that was the part a test could have caught.

Also note the branch was rebased onto the newer `master` (#4912) — I rebuilt the stack on top of that rather than force-pushing over it, so #4916 and #4917 moved too.

PR #4920, opened 2026-08-16
fix(logs): correct the alert row, and badge the events behind it

amir20:
Started as a one-line CSS fix and grew into the corrections that came out of
actually running the feature. Pairs with amir20/doligence#338.

### The row was painted red by a global rule

`LogLevel.vue` carries a second `<style>` block with **no `scoped` attribute**:

```css
[data-level="error"], [data-level="fatal"] { @apply !bg-red; }
```

That is global, and `!` compiles to `!important`, so anything in the app
carrying `data-level="error"` is filled red unconditionally. The alert row used
`data-level` for its own accent and could not win the cascade at any
specificity — which is why three separate attempts at changing its colour did
nothing. Both components now use `data-alert-level`.

`AlertGutter` had the same latent bug: its info ticks would have rendered green.

> `data-level` is effectively a reserved global attribute in this app and
> nothing says so. Scoping that block to its badge would remove the trap —
> deliberately out of scope here.

### The row now matches the agreed design

- one block: chip, headline and counts on top, actions right-aligned beneath
- red rail and chip, **no fill** — colour where the eye finds it fastest
- `font-sans`, not the log list's monospace: this is prose *about* the logs
- aligned with the log message column (log rows put a level dot before their
  message, so the slot's left edge is 18px short of where log text starts)
- unknown/empty levels read as error — `alerts.level` defaults to `''`, so the
  neutral fallback was the case that showed up most

### Alerts now appear on the window you open on

They only ever arrived through `loadOlderLogs`, so an alert that fired in the
window the viewer *opens* on never appeared — the common case, since "something
just happened" is usually why the container got opened.

That window asks for follow-up anchors, unlike scrollback: an incident that
started an hour ago and is still firing has no *origin* in the current window,
so origins-only rendered nothing precisely when the user was looking because
something is happening now.

### Every matched event is now visible

Cloud returns the individual events behind each alert (amir20/doligence#338),
and they are shown as a **badge on the log line itself** rather than as rows —
the event *is* that line, so a separate marker would duplicate what is on
screen, and during a storm would put one between every pair of lines.
`bell-alert` where a notification was sent, `bell-off` where it was held back.

Matching is by log id, the FNV-32a hash both sides stamp on the same raw line.
**Metric and event notifications carry no line, so they are skipped rather than
guessed at from a timestamp — they remain invisible in the stream and need
their own answer.**

### The scroll gutter is gone

It positioned ticks against *loaded* content, and the viewer loads
incrementally — every tick moved as scrollback arrived, and an alert outside the
range had no tick at all. A marker pointing at the wrong place is worse than
none.

### Translations

The alert strings existed only in `en.yml`, so every other locale rendered the
raw key (`LABEL.ALERT`). All twelve are now in all sixteen locales, following
each file's existing `container:` terminology.

### Verification

Driven with Playwright against the running stack rather than read off compiled
CSS — which is what let the `!important` rule hide for three rounds:

| | before | after |
| --- | --- | --- |
| row `background-color` | `oklch(0.64 0.218 28.85)` | `rgba(0, 0, 0, 0)` |
| background rules matching the row | the global `!important` one | none |
| alerts after a plain page load | 0 | 1 |
| badges on matched lines | 0 | 9 |

`typecheck`, `build`, `go build`, `go test ./internal/...` all pass. 153 UI tests
pass; the 3 `EventSource.spec.ts` snapshot failures are pre-existing on `master`
and were deliberately not refreshed, since `-u` would also rewrite them.


amir20:
Thanks — went through all of these. What changed since the reviews:

**Fixed: stale `buildTicks` auto-import.** Correct catch — `assets/auto-imports.d.ts` still declared it after `gutter.ts` was deleted. Removed, and confirmed a fresh build doesn't reintroduce it. (It didn't fail `pnpm typecheck` locally, so the generator's `declare global` block was tolerating it — but it was dead either way.)

**Moot: `AlertGutter.vue` dead selectors / missing tick test.** The gutter is gone entirely. It positioned ticks against *loaded* content while the viewer loads incrementally, so every tick moved as scrollback arrived and an alert outside the range had no tick at all. You were both right that those `fatal`/`critical`/`severe`/`unknown` branches could never match — `level` only ever emits `error`/`warn`/`info`.

**Added: regression guard for the CSS-collision class of bug.** Fair point that the specs assert on attributes, not computed styles, so they'd never have caught the original problem. `does not carry the globally-styled data-level attribute` now pins the *trigger* — the row must not carry `data-level`, since `LogLevel.vue`'s unscoped `!bg-red` paints anything that does. jsdom can't evaluate the cascade, so the attribute is the testable part; it's also verified against real computed styles in the browser.

**Checked: `proto/cloud/cloud.pb.go` in sync.** Regenerated with the same `protoc` invocation `main.go` declares, not hand-edited.

**Noted, not done: `summary = 17` out of sequence.** Harmless, and renumbering a published field would be worse.

**Noted, not done: scoping `LogLevel.vue`'s unscoped block.** Still worth a follow-up — `data-level` is effectively reserved app-wide with nothing saying so. Deliberately out of scope so this PR doesn't restyle unrelated components.

Also since the reviews, on feedback from testing:
- the alert row is now **one block per incident** — follow-up anchors are dropped at the merge, and everything after the alert shows through per-line badges instead
- **every matched event is badged** on the log line it matched (`bell-alert` sent, `bell-off` held back)
- alerts now appear on the window the viewer **opens** on, not only on scrollback
- all twelve strings translated across all sixteen locales
- **the suite is green**: 158/158. The three `EventSource` snapshots were failing on `master` before this branch; adding a scoped style to `LogItem.vue` stamps `data-v-*` into that captured markup, so they're refreshed here — the diff is scope attributes only.

PR #4951, opened 2026-08-26
fix(security)!: scope cloud search, alerts and debug to user labels, add notifications role

amir20:
Stacked on #4950. Audit of every label-sensitive path in `internal/web` turned up four more spots that ignore per-user container filters.

- `debug.go` listed every container on every host with an empty filter. Now uses `h.resolveLabels(r)`. Only mounted at debug level, but same bug.
- `GET /api/cloud/search/logs` returned hits with full log lines for any container in the instance. Hits outside the caller's scope are dropped.
- `GET /api/cloud/alerts` took `containerIds` straight from the client with no ACL check. Ids are intersected with the caller's scope, and an all-out-of-scope request returns empty without hitting cloud.
- Notification rules and dispatchers are instance-wide. Any authenticated user could write a rule matching containers their filter hides, aim it at a webhook they control, and read those log lines, and `listDispatchers` handed out every destination including its secrets.

For notifications this adds a `notifications` role next to `shell`/`actions`/`download`. It's part of `all`, so users with no `roles:` line keep access exactly as today. Rules and dispatchers 403 without it, and the bell is hidden via a new `enableNotifications` config flag.

The role does not make rules label-scoped, it just decides who can write them. Docs carry a warning that granting it to a filtered user lets them reach containers their filter hides. Per-rule scoping (owner + filter on the subscription, carried to agents in the proto) is a follow-up.

Two new helpers next to `resolveLabels`: `restrictedUser` (per-user filter, ignores the global `--filter` since that applies to everyone) and `visibleContainerIDs`.

Tests cover all four, including that cloud is never queried when every requested container is out of scope, and that unset roles still get notifications.

### DIFF
--- a/assets/components/Links.vue
+++ b/assets/components/Links.vue
@@ -4,6 +4,7 @@
     <Announcements />
 
     <router-link
+      v-if="config.enableNotifications"
       :to="{ name: '/notifications' }"
       :aria-label="$t('title.notifications')"
       data-testid="notifications"
--- a/assets/stores/config.ts
+++ b/assets/stores/config.ts
@@ -15,6 +15,7 @@ export interface Config {
   enableActions: boolean;
   enableShell: boolean;
   enableDownload: boolean;
+  enableNotifications: boolean;
   disableAvatars: boolean;
   releaseCheckMode: "automatic" | "manual";
   user?: {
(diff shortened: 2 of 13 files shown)


amir20:
Since this could break some folks, I am adding `!` to it.

PR #4957, opened 2026-08-27
fix(ui): prevent mobile search from jumping

PattyRich:
## Summary

Prevent the mobile log viewer from auto-scrolling while search results are being updated. This keeps the search field visible as matching rows populate.

Desktop behavior is unchanged.

## Testing

Verified manually on the mobile layout.


PattyRich:
Here is a recording of the mobile search behavior before and after.


https://github.com/user-attachments/assets/6b8e3891-6f20-4392-8b75-83b6ba5635e8

amir20:
I love the video. I wish everybody else did that.

PR #4961, opened 2026-08-28
feat(ui): move status icon next to container name

amir20:
Status icon moves to the left of the container name and the status column becomes plain text.

Containers without a healthcheck now get an icon too, colored by state (running green, exited/dead red, paused/restarting/starting orange, created/deleted muted), so every row lines up.

https://claude.ai/code/session_01RxkKc3YHhUX9hFgti8Qj86


assets/components/ContainerViewer/ContainerStatusIcon.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="inline-flex size-4" :data-status="status" :title="status">
+    <cil:check-circle v-if="health === 'healthy'" />
+    <cil:x-circle v-else-if="health === 'unhealthy'" />
+    <cil:media-pause v-else-if="state === 'paused'" />
+    <cil:circle v-else />
```
amir20: this seems like an empty circle. does that make sense?

amir20: good catch, an empty ring reads like nothing. switched to a filled dot (mdi:circle-medium) so running/created/exited without a healthcheck show a solid status dot, colored by state.

PR #4971, opened 2026-08-31
feat: add network topology and dependency map views

always1ov:
Adds two pages that visualize live containers. /topology draws containers as nodes linked by the docker networks they share, with a one-shot force layout, pan/zoom, hover highlighting and a details card. /dependency-map groups compose projects into cards, draws depends_on edges resolved from compose labels and dashed edges between stacks sharing a network.

The backend reports each container's network names in a new Networks field, populated from NetworkSettings in both list and inspect paths and carried over the agent proto (field 27). GET /api/topology exposes the same graph as JSON: containers with networks, resolved depends_on targets, and network/stack membership.

Default bridge, host, none and ingress networks are shown as standalone nodes without links since they carry no grouping signal. Networks with more than four members render hub-and-chain instead of a full mesh to stay readable.


amir20:
This is a lot of code. I'll have to review it more. 

Some thoughts: 

What problem is this solving? 
What does it look like for you? For me it looks pretty confusing and not that useful. 
The top of the of the page is prime space. This isn't for everybody so I wouldn't agree it makes sense there. 
How does this help investigate logs, issues or whatever?

PR #4976, opened 2026-09-01
feat: show app icons for well known container images

amir20:
Matches container images to their project logo and shows it next to the container name in the sidebar, the container table, and the command palette. Most people running Dozzle are running an arr stack or something like it, and 30 containers are much faster to scan by logo than by name.

## Icons are bundled, never fetched

Vendored from [homarr-labs/dashboard-icons](https://github.com/homarr-labs/dashboard-icons) into `assets/icons/apps` and embedded in the binary. Requesting `sonarr.svg` from a CDN would tell a third party what a user is running, and it would break air-gapped installs.

251 files, 542KB. SVG where upstream has one, WebP downscaled to 64px otherwise, plus a few oversized SVGs rasterized (jdownloader was 330KB on its own). `assetsInlineLimit` keeps the directory out of the entry chunk so each icon stays a separately fetched, hashed asset and the browser only pulls what is on screen.

## Status moved onto the logo

Two glyphs side by side made the sidebar noisy, so `ContainerStatusIcon` is now `ContainerIcon` and the status is a corner badge:

- healthy: green check
- unhealthy: red exclamation
- otherwise: dot in the state colour
- not running: logo goes grayscale and dim

Containers with no matching icon fall back to the old standalone glyph, so nothing regresses for images that are not recognized.

## Matching

Last path segment of the image, ignoring registry, tag, and digest. `sonarr`, `linuxserver/sonarr:latest`, `lscr.io/linuxserver/sonarr`, and `ghcr.io/hotio/sonarr@sha256:...` all resolve to Sonarr. When the segment is generic it falls back to the namespace, which is what rescues `ghcr.io/goauthentik/server`. An alias table covers the rest (`postgres` to postgresql, `plexinc/pms-docker` to plex, and so on).

`dev.dozzle.icon` overrides the guess, or set it to `none` to hide the icon for one container.

The feature is behind Settings, Options, Show app icons, defaulted on.

## Testing

Verified against 44 real image names, 43 resolve correctly (the misses are apps with no bundled icon). Ran a fake arr stack of 39 containers locally and checked running, healthy, unhealthy, stopped, and unrecognized states render correctly.

`pnpm test` 238 passing, `pnpm typecheck` and `go build ./...` clean.

Docs at `docs/guide/app-icons.md`.

https://claude.ai/code/session_01G4tza33QFYH98eidfoccyf


svenjacobs:
Really cool feature! Could you also add the icon for `copyparty`, please?
It's available [here](https://github.com/homarr-labs/dashboard-icons/blob/1cdb6d737c3623705109fb8f4aee890adab58d3b/svg/copyparty.svg) but for some reason wasn't copied into Dozzle.

amir20:
Thanks! I like this feature too. Send a PR. Instructions are at https://github.com/amir20/dozzle/tree/master/assets/icons

PR #4983, opened 2026-09-02
fix(auth): resolve roles from users.yml instead of the JWT claim

amir20:
#4979 only covered tokens with no `roles` claim. A session minted between the roles feature (5f24657b) and the cloud role (3794a3b6) has a claim, it just predates the `Cloud` bit, so it carries `roles: 30` and parses fine. The `roles := All` default never applies and cloud stays hidden. The signing key digest hashes users.yml, and adding a role in code does not touch users.yml, so the key does not roll either. Only logging out fixes it.

Same thing happens for any future role bit, and for a `roles:` or `filter:` edit in users.yml on a live session.

Simple auth now resolves the user from users.yml on every request and puts them in the request context, the same way proxy auth already does. The token proves identity, nothing else.

- `UserFromContext` drops the claims fallback, so roles have one source of truth. That also closes the path where a still-valid token for a user removed from users.yml got `All`.
- `UserDatabase.Find` reloads users.yml in place and now runs per request, so it is behind a mutex.
- Replaced the JWT-claims tests with middleware tests: stale mask picks up cloud, revoked role applies to a live session, token with no roles claim, unknown user, no token.

Role and filter edits in users.yml now take effect on the next request rather than at next login.

`go build ./...`, `go vet ./internal/...` and `go test -race ./...` pass.

https://claude.ai/code/session_01X8fhaUTBkCruxzHioZSCQn

### DIFF
--- a/internal/auth/simple.go
+++ b/internal/auth/simple.go
@@ -47,18 +47,31 @@ func NewSimpleAuth(userDatabase UserDatabase, ttl time.Duration) *simpleAuthCont
 	}
 }
 
+// find and findByPassword copy the user out of the database rather than handing
+// back the pointer into UserDatabase.Users, so a caller reading it after the lock
+// is released cannot race a reload of users.yml.
 func (a *simpleAuthContext) find(username string) *User {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	return a.UserDatabase.Find(username)
+	return copyUser(a.UserDatabase.Find(username))
 }
 
 func (a *simpleAuthContext) findByPassword(username, password string) *User {
 	a.mu.Lock()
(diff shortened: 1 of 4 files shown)


amir20:
pushed e6fca162:

- `find`/`findByPassword` now return a copy, so nothing holds a pointer into `UserDatabase.Users` after the lock drops. no longer depends on `readFileIfChanged` swapping the map wholesale.
- resolved user carries `Filter` and `RolesConfigured` from the db entry. checked `User.Filter` has no post-auth reader (only `CreateToken` and `generate_command`), `ContainerLabels` is what downstream uses, but the struct may as well match users.yml.
- added tests for filter resolution and for the copy.

PR #4985, opened 2026-09-02
feat(icons): add copyparty, atuin and ddclient app icons

svenjacobs:
Adds three icons to `assets/icons/apps/`, vendored from
[homarr-labs/dashboard-icons](https://github.com/homarr-labs/dashboard-icons) per the
process in `assets/icons/README.md`:

- `copyparty.svg`
- `atuin.svg` + `atuin-light.svg` (upstream ships a dark-background variant)
- `ddclient.svg`

All three slugs match their image names, so no `ALIASES` entries in `appIcons.ts` are
needed. Resolution verified for the images people actually run:

| image | resolves via |
| --- | --- |
| `ghcr.io/atuinsh/atuin` | name |
| `linuxserver/ddclient`, `ghcr.io/ddclient/ddclient` | name |
| `copyparty/ac` | namespace fallback |

`ghcr.io/9001/copyparty-ac` (and the `-min`, `-iv`, `-dj` flavors) won't match, since the
slug sits under the numeric namespace and the suffix isn't in `SUFFIX`. Left it alone
rather than adding five aliases for it. Happy to add them if you'd rather cover that form.

Existing `appIcons.spec.ts` suite passes (25 tests).


svenjacobs:
I assume the Claude Review action fails because the token is not available in CI environments of forks.

Issue #4992, opened 2026-09-03
Agents don't work in Cloud


amir20:
### Discussed in https://github.com/amir20/dozzle/discussions/4986

<div type='discussions-op-text'>

<sup>Originally posted by **chrisgla** September  2, 2026</sup>
I have one host with an agent/hub in the host docker instance and a second agent in home assistant docker suite on the same host.

Both agents appear in local hub and under the instance in cloud dashboard, and I can see all logs in cloud - but I cant find how to see the second agent summary in overview?</div>

amir20:
This is fixed now with 10.9.1

PR #4993, opened 2026-09-03
fix(cloud): report remote agents to Dozzle Cloud

amir20:
A hub with `--remote-agent` shows only its own Docker host in Dozzle Cloud. The agent's host, containers, logs and metrics never arrive — not hidden in the UI, never sent.

## Cause

`newLocalCloudHostService` scoped everything the cloud client can see to `LocalClientServices()`, which filters to `*DockerClientService`. An agent is an `agentService`, so it was dropped. That one wrapper backs `list_hosts`, `find_containers`, `list_running_containers`, the log streamer and the stats streamer, so an agent was invisible to all of them.

`--remote-host` kept working only because a `tcp://` daemon happens to be a `DockerClientService` — that asymmetry is the tell.

The scoping exists for swarm, where every replica discovers every peer and reporting the fleet from each would multiply hosts, and duplicate log and stat ingestion, by replica count. Outside swarm there is exactly one Dozzle holding the fleet: `--remote-agent` and `--remote-host` endpoints are configured on it and on nothing else.

## Fix

Keep the local scope in swarm mode only. Elsewhere the cloud client gets the whole fleet, via a new `MultiHostService.ClientServices(retry bool)` that does not filter by client type.

Two related fixes while in here:

- The fleet is read per call instead of snapshotted at startup, and log/stat subscriptions extend to hosts that become available later. An agent that is down at boot no longer stays invisible until a restart.
- Retrying unreachable agents is opt-in per call site. The periodic fan-outs pay for it; `FindContainer` does not, because it runs once per log reader and a run of them against a black-holed agent would each wait out the dial timeout.

## Tests

`main_test.go`: agents present in server mode, absent in swarm, and a late-joining agent appearing without a restart.

## Affected versions

The scoping landed in c5c301f (v10.5.1), so anyone on an older Dozzle never had this.

Fixes #4986 https://github.com/amir20/dozzle/issues/4992

https://claude.ai/code/session_019YoEbYyS65nfGsazg3egEd

### DIFF
--- a/internal/support/docker/multi_host_service.go
+++ b/internal/support/docker/multi_host_service.go
@@ -188,6 +188,23 @@ func (m *MultiHostService) LocalClientServices() []container_support.ClientServi
 	return m.manager.LocalClientServices()
 }
 
+// ClientServices returns every client this Dozzle serves — the local docker
+// daemon, --remote-host daemons and --remote-agent agents alike. Unlike
+// LocalClientServices it does not filter by client type, so a caller that needs
+// the whole fleet (the cloud client) sees agents too.
+//
+// retry re-dials agents that were unreachable the last time they were tried, so
+// one that came up since joins the returned set. It costs a connection attempt
+// per still-unreachable agent, up to the configured timeout each — pass it on
+// the periodic fan-out calls, not on per-container lookups.
+func (m *MultiHostService) ClientServices(retry bool) []container_support.ClientService {
+	if !retry {
+		return m.manager.List()
+	}
+	services, _ := m.manager.RetryAndList()
+	return services
+}
+
 func (m *MultiHostService) TotalClients() int {
 	return len(m.manager.List())
 }
--- a/internal/support/docker/retriable_client_manager.go
+++ b/internal/support/docker/retriable_client_manager.go
@@ -18,12 +18,23 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+// hostSubscriber is one registered listener for "a host became available".
+//
+// Subscriptions are keyed by this pointer rather than by their context: one
+// caller may open several subscriptions under a single context (the cloud
+// client's log and stat streamers share one stream lifetime), and keying by
+// context would let the second registration silently evict the first.
+type hostSubscriber struct {
+	ctx     context.Context
+	channel chan<- container.Host
+}
+
 type RetriableClientManager struct {
 	clients      map[string]container_support.ClientService
 	failedAgents []string
 	certs        tls.Certificate
 	mu           sync.RWMutex
-	subscribers  *xsync.Map[context.Context, chan<- container.Host]
+	subscribers  *xsync.Map[*hostSubscriber, struct{}]
 	timeout      time.Duration
 }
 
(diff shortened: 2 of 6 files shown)


amir20:
@chrisgla can you try this `amir20/dozzle:pr-4993`? I have tested it and it works. I'll probably release in a couple days when I have other fixes in.

amir20:
All three addressed in 75c38f2.

**Subscriber map key collision** — correct, and it was a live bug the moment both streamers watched for hosts. Fixed at the root rather than by handing `watchNewServices` a derived context: `subscribers` in both `RetriableClientManager` and `SwarmClientManager` is now keyed by the subscription itself, not by its context. Keying by context only ever held because every caller happened to pass a distinct one, and nothing said so. `retriable_client_manager_test.go` covers two subscriptions sharing one context, and `TestCloudHostService_SubscribesLateJoiningAgents` covers the subscription path the existing late-join test missed.

**Transient `Host()` failure** — also correct. `hostID` now takes the id whether or not the call errored: an agent only enters the fleet after identifying itself once, and `agentService.Host` returns its last known host alongside the error, so a blip no longer leaves a connected agent unstamped. Separately, re-attach now runs on a timer as well as on the unreachable-to-reachable edge, so anything skipped for any reason recovers instead of staying skipped for the life of the connection — that removes the dependence on the notification path firing at all.

**Dead `l.hs == nil` check** — right, removed.

Also dropped the unused test field staticcheck flagged. `-race` suite and staticcheck are clean.

chrisgla:
> @chrisgla can you try this `amir20/dozzle:pr-4993`? I have tested it and it works. I'll probably release in a couple days when I have other fixes in.

Tested pr-4993 on my hub (non-swarm, 3 remote agents), (the hub UI) is fixed — all 4 hosts and 72 containers show correctly with full logs/stats. cloud.dozzle.dev is still broken though: the Overview page only ever shows 1 host / 23 containers (matches the hub's own local count), and the Instances page briefly shows correct per-agent container counts before they drop back to 0. Debug logs show list_hosts only gets called twice at hub startup (two concurrent request IDs racing) and nothing after — no periodic cloud tool calls, no errors, no reconnect/eviction messages in between. Looks like cloud is either caching one of the two racing startup responses or the periodic refresh into cloud isn't happening/silently returning empty.

amir20:
Hi @chrisgla,

I think I need a little more detail. I can't see your data for privacy reasons. But I can see what you have connected. I am not sure what you mean by still broken. Here are some things I think you should check: 

1. I noticed only one Dozzle has been updated to `pr-4993`. `v10.9.1` is released now. You should update _all_ dozzle instance to latest. 
2. "Overview page only ever shows 1 host / 23 containers" -- Where is that? I don't show any host info on the overview page. 
3. There is a bit of caching, so you maybe right about it. But would it be possible provide screenshots with arrows pointing to what you think is missing. 


Thanks for helping me while I make Cloud better.  I fixed a lot of bugs yesterday in 10.9.1 so that should've helped. 

edit: To make it easier for us to debug, I have added more details to the instances page. Make sure to update to `10.9.2`

amir20:
Also follow up question. After looking at what's connected, something doesn't make sense. Do you have one UI instance and 3 agents? That would make more sense.

Issue #4988, opened 2026-09-02
Bundle app icons: django, powerdns, rspamd


codingjoe:
... at least Django is somewhat popular 😉

amir20:
Send a PR. :)

amir20:
Closing. PRs welcomed.

PR #5001, opened 2026-09-05
feat(icons): bundle web framework app icons

codingjoe:
Ref: #4988


amir20:
@codingjoe would this actually work? Most apps aren't called by their framework.

codingjoe:
I guess that depends on how you build an name your container. But more commonly people ( and me) would set a label in their compose file that choose a framework. Combinations like DRF and Next or FastAPI and Next are very common. So, it be nice to have different icons for your backend and frontend containers. 

That was my line of thought. I actually have 10 different app containers + replication in my current project and it's getting hard to tell them appart 😅

amir20:
Good point. I forgot the support for labels.

PR #5004, opened 2026-09-05
feat(ui): dev.dozzle.url label to link a container to its own web UI

### LINKED ISSUE #5002: feat: dev.dozzle.url label to link a container to its own web UI

Most containers I watch in Dozzle also serve a web UI — Grafana, Portainer, a database console. Dozzle is already the page I have open to see what they're doing, so it's the natural place to jump to them from, but there's no way to get there. I keep a separate dashboard open purely to hold those links.

A `dev.dozzle.url` label would close that: put the address in the compose file, and Dozzle offers it as a link on that container.

```yaml
services:
  grafana:
    image: grafana/grafana
    labels:
      dev.dozzle.url: https://grafana.example.com
```

It's opt-in and explicit — no port scanning, no guessing at what a container exposes. Dozzle only shows what I told it to, which sidesteps the auto-discovery concern raised in #3285.

Anywhere that doesn't steal the plain click, which opens the logs. A few placements, roughly in order of how much UI they add:

1. **A column in the container table** — a short "Link" or "Open" column at either end of the row, rendered only for containers that carry the label, and hidden entirely when no container has one. Most discoverable spot, and it doesn't touch any existing behaviour.
2. **A small external-link icon beside the container name**, in the sidebar and the table, next to where `dev.dozzle.icon` already renders.
3. **Modifier-click** — Cmd+click on macOS, Ctrl+click elsewhere — opens the URL in a new tab, matching browser convention. No new UI at all, but undiscoverable on its own, so best as a companion to 1 or 2 rather than the only affordance.

One implementation note: worth restricting the value to `http`/`https` before rendering, so a label can't slip in a `javascript:` URL, and putting `rel="noopener noreferrer"` on the anchor.

- A separate dashboard (Homepage, Homarr) just for the links. Works, but it's a second file to keep in sync with compose, for something I only want while I'm already in Dozzle.
- #3285 (exposed ports), closed as not planned. This is deliberately narrower — no discovery, no port list, just a link the operator wrote down.
- Bookmarks. Fine until a container moves host or port.

amir20:
Closes #5002

Opt-in `dev.dozzle.url` label. When it's set, Dozzle shows a link to the container's own web UI.

```yaml
services:
  grafana:
    image: grafana/grafana
    labels:
      - dev.dozzle.url=https://grafana.example.com
```

The link shows up next to the container name in the container table, in the sidebar hover popup, and in the container title bar. Only absolute http/https values are honored, so a label can't slip in `javascript:`, and every link gets `target="_blank" rel="noopener noreferrer"`. The plain click still opens logs.

No auto detection. Published ports aren't reachable URLs often enough for guessing to be worth it, which is the same reason #3285 was closed.

## Discoverability

A container that publishes a port and has no label yet gets a small icon in the title bar. It opens a compose snippet prefilled with the published port, ready to copy. The guess only ever prefills text you edit, it never becomes a rendered link, so getting it wrong costs nothing. "Don't show again" hides it everywhere and syncs to the profile.

The popover is a reusable `DiscoveryHint` component (title, body slot, optional docs link, dismiss) for the next feature that needs teaching.

## Hover popup

Reworked while it was being touched. It now shows the image tag, health, CPU/memory sparklines, published ports, group, and host, on a fixed width instead of stretching to fit. It also stays open when the pointer moves onto it, and switching rows swaps panels immediately instead of overlapping.

## Notes

- `Ports` on the container model now serializes to JSON. It already crossed the agent proto, it just wasn't exposed to the frontend.
- All 16 locales updated.
- Docs: `docs/guide/container-links.md`.

Modifier-click (option 3 in the issue) is not included. Cmd+click on a row is already "open the logs in a new tab", and the icon covers the same ground.

https://claude.ai/code/session_01MEjLBJYCRJ2Pqpf4NnrAK4


AmerJadid:
Tried it — ran `amir20/dozzle:pr-5004` against a throwaway read-only socket proxy with a handful of probe containers. It's good, and the two things I was most worried about you got right: the hint only ever prefills a copyable snippet instead of rendering a guessed URL as a live link, and the scheme check is a real allowlist rather than a denylist.

Threw some hostile label values at it. `javascript:`, `data:`, `vbscript:`, `file:`, `ws:`, mixed-case `JaVaScRiPt:`, a tab inside the scheme and whitespace padding are all rejected, and `//evil.example.com` throws in `new URL` so it's dropped too. `rel="noopener noreferrer"` is there.

Three small notes:

**1. `url` returns `raw` rather than the parsed URL.** A homograph therefore survives into both the href and the tooltip:

```
label             https://exаmple.com      ← Cyrillic а
new URL(raw).href https://xn--exmple-4nf.com/
```

Returning `u.href` normalizes it to punycode in both places, which makes the tooltip an actual defence rather than one that reads `example.com`. The same getter could expose `u.host` for display — that drops userinfo as a side effect, so `https://admin:hunter2@example.com/panel` would stop showing credentials to everyone who can see the container.

**2. The snippet drops the bind address.** `portMappings` takes `from.split(":").pop()`, so `127.0.0.1:18105->80/tcp` and `0.0.0.0:18105->80/tcp` both suggest `http://<hostname>:18105`. For a loopback-bound port viewed from a remote browser, that URL can't work — and binding published ports to `127.0.0.1` is a pretty common way to run things behind a proxy. Not a security issue, since it's a snippet the user edits before it means anything, just a suggestion that's wrong in that setup. Skipping `127.0.0.1`/`::1` bindings when the page isn't served from that host would cover it, or a line in the docs. (The IPv6 dedup and the udp filter are nice touches.)

**3. `Ports` went from `json:"-"` to `json:"ports,omitempty"`.** The published-port list now reaches every client, which is the data #3285 asked for. Your design needs the actual numbers to build the snippet so I don't think there's a way around it — just flagging that it's a decision rather than a side effect, since you'd closed that one.

Nothing here blocks it from my side. Happy with the shape, and thanks for turning it around this fast.

amir20:
OK, any improvements will be in follow up PRs.

PR #5039, opened 2026-09-07
fix(logs): stop merged log stream from deadlocking the container store

amir20:
Production instance stopped picking up new containers. Two containers redeployed at 18:47 were still missing four hours later, with `container not found` logged twice a second the whole time. A goroutine dump from the wedged process pointed at the merged log stream handler.

## The deadlock

`streamLogsForContainers` has a select loop that is the only reader of its own `events` channel (cap 1), and one of its cases also writes to it:

```go
events := make(chan *container.ContainerEvent, 1)
...
case c := <-newContainers:
    events <- &container.ContainerEvent{...}   // waits on this same goroutine
case event := <-events:
```

The other producer is `streamLogs`, which pushes `container-stopped` when a container's log stream ends. If that slot is occupied when a container starts, the loop blocks on a channel only it can drain. A redeploy is exactly "one container stops while another starts", so it happens under normal use and nowhere else, which is why it looked random.

## Why one browser tab took the host down

```
handler loop stops draining newContainers (unbuffered)
  -> MultiHostService forwarder blocks (ctx is the request, still alive)
  -> ContainerStore.addContainer blocks broadcasting to subscribers
  -> the whole per-host store loop stops
```

That loop also drains the Docker event stream and stats. Docker does not hold events for a consumer that stopped reading, so `create` and `start` are dropped while it is stuck, and those are the only path a container has into the store. Anything that starts during the freeze is invisible until the process restarts. It also explains the 255 second stats gap in the logs and the 257 samples that flushed in one second when the browser finally disconnected.

## Changes

* `logs.go`: emit `container-started` straight to the SSE writer, which the loop already owns.
* `container_store.go`: fan out with a bounded send (250ms) instead of a blocking one. A subscriber that will not read loses that message and gets logged, instead of stopping every client on the host.
* ctx-guard the sends that leak a goroutine and its Docker stream once the consumer is gone: `StreamLogs` (docker, k8s), `ContainerEvents` (docker, k8s), agent client log and event streams. The dump had one of these parked on a send for 26 hours.

## Tests

Two regression tests, both verified to fail on master:

* `Test_handler_streamHostLogs_container_started_while_event_buffered` drives the exact ordering and times out on master.
* `TestContainerStore_wedgedSubscriberDoesNotStallStore` subscribes two readers that never read and asserts later containers still reach the store.

Full suite passes with `-race`.

https://claude.ai/code/session_017cuUsCNfGQWDNrw1k2mp1W

### DIFF
--- a/internal/agent/client.go
+++ b/internal/agent/client.go
@@ -207,7 +207,8 @@ func sendLogs(stream pb.AgentService_StreamLogsClient, events chan<- *container.
 			continue
 		}
 
-		events <- &container.LogEvent{
+		select {
+		case events <- &container.LogEvent{
 			Id:          resp.Event.Id,
 			ContainerID: resp.Event.ContainerId,
 			Message:     message,
@@ -216,6 +217,9 @@ func sendLogs(stream pb.AgentService_StreamLogsClient, events chan<- *container.
 			Level:       resp.Event.Level,
 			Stream:      resp.Event.Stream,
 			RawMessage:  resp.Event.RawMessage,
+		}:
+		case <-stream.Context().Done():
+			return stream.Context().Err()
 		}
 	}
 }
@@ -304,7 +308,11 @@ func (c *Client) StreamEvents(ctx context.Context, events chan<- container.Conta
 			c := container.FromProto(resp.Event.Container)
 			evt.Container = &c
 		}
-		events <- evt
+		select {
+		case events <- evt:
+		case <-stream.Context().Done():
+			return stream.Context().Err()
+		}
 	}
 }
 
--- a/internal/agent/client.go
+++ b/internal/agent/client.go
@@ -272,7 +272,8 @@ func (c *Client) StreamStats(ctx context.Context, stats chan<- container.Contain
 			return rpcErrToErr(err)
 		}
 
-		stats <- container.ContainerStat{
+		select {
+		case stats <- container.ContainerStat{
 			CPUPercent:     resp.Stat.CpuPercent,
 			MemoryPercent:  resp.Stat.MemoryPercent,
 			MemoryUsage:    resp.Stat.MemoryUsage,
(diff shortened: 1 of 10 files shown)


amir20:
All three addressed in 3e12f4f9.

- `StreamStats` and `StreamNewContainers` now guard on `stream.Context()` like `sendLogs` and `StreamEvents`.
- Fan-out timeout is now shared per broadcast instead of per subscriber. Five wedged readers cost the loop one 250ms budget between them, not five. Lazily started, so a fan-out where everyone is reading allocates nothing. Covered by `TestContainerStore_broadcastBudgetIsShared`.

Audit turned up three more of the same class:

- `agent/client.go` `UpdateContainer` progress send.
- `DockerClientService.UpdateContainer` pushes progress into a channel a request owns. The SSE handler in `actions.go` returns the moment a write to the client fails, and the agent stream ends with its RPC, either of which strands the update goroutine partway through the stop/remove/create/start sequence it is running. All 13 sends go through a ctx-guarded helper now.
- `EventGenerator.consumeReader`. `processBuffer` stops draining as soon as `emit` sees the context end, so once the 100-deep buffer fills, `consumeReader` parks forever holding its reader. This is the generator behind every log stream.

Also scanned every function that both reads and writes the same channel, looking for more of the self-deadlock. `streamLogsForContainers` was the only one.

amir20:
Good catch on the ordering, and it does matter: `eventStreams.ts` sorts the buffer only on the first flush (`flushNow`, the `initial` branch), everything after that renders by arrival. A start listed above the stop that preceded it would have been visible in the live view during a redeploy, which is exactly when both events fire.

Fixed in 765275f8 by queueing the started event through `events` from its own goroutine instead of writing it straight to the SSE writer. The select loop still never sends on the channel it drains, so the deadlock stays fixed, and both events share one FIFO again so they keep their order.

Agreed on `g.Errors`, cap-1 and written once on the way out, so it cannot wedge. Left alone.

PR #5043, opened 2026-09-07
feat(auth): sign in with GitHub and OIDC 🥳

amir20:
Adds "Sign in with GitHub" and generic OIDC as a second way to authenticate into the existing `simple` provider. It is not a fourth auth provider. `users.yml` stays the user list and Dozzle still serves its own login page.

## How it works

`users.yml` is the allowlist. An external account that no entry links to cannot sign in, and no account is ever created automatically. Roles and filters resolve per request from `users.yml` exactly as they do for password login, so nothing downstream of the token changed.

One new optional field, one existing field made optional:

```yaml
users:
  amir:
    email: amir@example.com # OIDC matches on this
    name: Amir
    password: $2a$11$... # now optional
    github: amir20 # new, GitHub matches on this
    roles: all
```

Password login keeps working alongside, which matters when an OAuth app breaks and you need to get in to fix it. The username and password fields hide themselves when nobody in `users.yml` has a password.

## Config

```
DOZZLE_AUTH_PROVIDER=simple            # github and google accepted as aliases
DOZZLE_AUTH_GITHUB_CLIENT_ID=...
DOZZLE_AUTH_GITHUB_CLIENT_SECRET=...
DOZZLE_AUTH_OIDC_ISSUER=https://id.example.com
DOZZLE_AUTH_OIDC_CLIENT_ID=...
DOZZLE_AUTH_OIDC_CLIENT_SECRET=...
DOZZLE_AUTH_OIDC_NAME=Pocket ID        # button label, defaults to SSO
```

Both client secrets also accept a `_FILE` counterpart for Docker secrets. Setting a variable and its `_FILE` together is an error, and a missing or empty file stops startup instead of quietly dropping the login button.

One generic OIDC implementation covers Google, Keycloak, Pocket ID, Zitadel and Authentik.

## Design notes

OAuth is an optional `OAuthAuthorizer` interface that `routes.go` type-asserts, rather than a new `AuthProvider` value. Branching on the provider would put the same user behind two role resolution paths. A third vendor touches no routing code.

GitHub matches on the login, OIDC matches on the verified email. A GitHub login is stable and always present while a GitHub email can be private or changed. OIDC has no stable human readable equivalent, so the verified email is the claim operators actually know. OIDC rejects a sign in when `email_verified` is false, since matching an unverified address would let anyone who can register at a permissive IdP claim an account by typing in someone else's email.

GitHub omits `redirect_uri` and lets the OAuth app decide. OIDC requires it, so it is derived from `Host` plus `X-Forwarded-Proto` and pinned in the state cookie so the token exchange replays the identical value. A spoofed Host fails the login rather than leaking a code, because the IdP only honors a `redirect_uri` on its registered list. No new base URL flag either way.

The `id_token` signature is not verified. The code is exchanged by this process directly against the endpoint named in a discovery document fetched over TLS from the configured issuer, authenticated with the client secret, and identity is then read from userinfo over that same channel, so the claims never pass through the browser. Discovery also rejects a document whose `issuer` disagrees with the configured one. Happy to add `jwx` verification if you'd rather have it.

Editing `users.yml` still rotates the JWT signing key and logs everyone out. `email` and `github` joined the key derivation so repointing a linked account rotates sessions the way a password change does.

Deliberately no org-wide or domain-wide allowlisting. `forward-proxy` with Authelia or Authentik is the right tool for that, and adding it here would break `users.yml` as the single source of truth.

## Also in here

- Fixes an open redirect on the login page. A request for `//evil.com` came back as a protocol relative `redirectUrl` that the login page would follow off origin. `redirectUrl` is also query encoded now, so a path containing `&` survives.
- `users.yml` validation returns errors instead of `log.Fatal`. Startup behavior is unchanged (`main.go` still fatals), but a bad edit no longer kills a running server mid reload.
- Splits `docs/guide/authentication.md` (590 lines) into an overview plus three sub pages. Security Considerations and Generating users.yml stay on the overview so existing inbound anchors keep working.
- Drops the SHA-256 password hash warning from the docs.
- Redesigns the login page and the profile dropdown. The old login markup put daisyUI's `input` and `floating-label` on the same element, which is what made the label sit on the border, and both inputs carried `autofocus`.
- 3 new i18n keys, translated into all 15 other locales. Docs mirrored into `de`, `es`, `fr` and `zh`.

## Testing

48 new Go tests covering state and PKCE mismatch, expired state, provider errors, users not in `users.yml`, case folding, unverified email, issuer mismatch, discovery caching and retry, redirect URI replay, and the `_FILE` cases.

`go build`, `go vet`, `go test -race`, `pnpm typecheck` and 281 vitest tests pass. `_FILE` was also verified end to end against a real binary for all four paths (reads the file, both set, missing file, empty file).

Every `simple.spec.ts` test passes, including the login and logout flows through the redesigned page. 8 unrelated tests failed locally (`agent`, `remote-host`, `mobile-stats`, and the 4 `visual` homepage snapshots) because two stray `dozzle-link-labeled` / `dozzle-link-unlabeled` containers on my Docker host show up in the container list. They are not in `docker-compose.yml`, and this branch touches neither the compose fixtures, the e2e specs, nor the side panel. CI should be clean.

No e2e for the OAuth flow itself. Mocking an IdP costs more than it catches.

https://claude.ai/code/session_01Ho652RzbWLdefrWSb4RNUc


amir20:
Went through the review comments. Two were real, one is not.

**Fixed: OIDC discovery failure sent the browser to a relative URL.** Confirmed by reproducing it against an issuer on a refused port. `AuthCodeURL` on a zero-value `Endpoint` produced `/api/auth/?client_id=...&state=...`, so the browser followed a relative URL back into Dozzle carrying the state and PKCE params instead of landing on `?error=oauth`.

`oauth2Config` now returns an error instead of a half-built config, so the failure cannot be swallowed. GitHub returns nil (its endpoints are constants); OIDC surfaces the discovery error and both handlers call `failLogin`. The config is also built before the state cookie is set, so a login that cannot start leaves nothing behind. Two regression tests cover it, driving `LoginHandler` and `CallbackHandler` rather than `discover()` directly.

**Fixed: stale `--auth-provider` help text.**

**Not a bug: the post-login redirect.** `req.URL.String()` does not already start with `/` here. The index handler is mounted behind `http.StripPrefix(base+"/")` (routes.go), so the path arrives with the prefix stripped and no leading slash, which is why the `"/" +` is there. Verified against a running binary:

```
GET /               ->  Location: /login?redirectUrl=%2F
GET /container/abc  ->  Location: /login?redirectUrl=%2Fcontainer%2Fabc
GET /settings       ->  Location: /login?redirectUrl=%2Fsettings
```

Deep links are preserved. Dropping the prefix as suggested would break them.

**On matching GitHub logins rather than the numeric id:** the risk is real and now documented as a warning. A rename frees the old handle and whoever registers it inherits that `users.yml` entry, so a rename has to be treated as an access change. Matching on `id` would close it but puts opaque numbers in a file people hand-edit, so it stays a login for now. If we want to close it properly, an optional `github_id:` alongside `github:` is the shape I would go for, as a follow-up.

amir20:
Followed up on both non-blocking nits from the last review. The first turned out to be more than a nit.

**`login.vue`'s redirect guard was a live open redirect.** The client-side check was a prefix test (starts with `/`, not `//`, not `/\`). `"/\t/evil.com"` passes all three, but the URL parser strips ASCII tabs and newlines, so the browser resolves it to `https://evil.com`:

```
"/\t/evil.com"  guard passes: true  -> resolves to: https://evil.com
"/\n/evil.com"  guard passes: true  -> resolves to: https://evil.com
"//evil.com"    guard passes: false -> resolves to: https://evil.com
```

So a crafted `?redirectUrl=/%09/evil.com` sent anyone who logged in with a password off-site. The Go side already rejected all three (`url.Parse` errors on control characters), so this was frontend-only, which is exactly the divergence the review flagged.

Resolving with the browser's own parser and comparing origins is the only check that agrees with what navigation actually does. The structural check still runs first so behaviour does not depend on whether a base path is configured. Pulled out of the SFC into `assets/composable/safeRedirect.ts` with 12 test cases covering tab, newline, carriage return, backslash, protocol-relative and absolute forms plus base handling, so the two implementations can be compared directly rather than drifting.

**GitHub `/user/emails`** now logs at debug when it fails. Still not fatal, since matching is on the login, but a missing `user:email` scope otherwise looks identical to a user with no verified address.

294 JS tests, 56 auth tests, `go vet` and `go test -race` clean. Rebased onto the latest master merge.

PR #5067, opened 2026-09-09
fix(logs): apply field toggles in service views

### LINKED ISSUE #5036: Field toggles have no effect on swarm mode service log lists

- [x] Completed

### How is Dozzle deployed?

Swarm Mode

### 📦 Dozzle version

10.9.0

```
services:
  monitoring:
    image: amir20/dozzle:latest
    environment:
      DOZZLE_MODE: swarm
    volumes:
    networks:
      - main_network
    deploy:
      mode: global
      update_config:
        order: stop-first
        failure_action: rollback
```

Hello,

in docker swarm mode we have two types of access to logs. 1 is by service name and the other is a direct container access.

If i have a service called `my-service` which has 3 replicas `my-service-1`, `my-service-2`, `my-service-3` i can configure the fields visible in log-list for the containers (1,2,3). However, in service-view where the logs of all three containers are listed together, the field toggles have no effect on the list.

Worth noting: My logs are in json format

Kind regards and thank you for the amazing work <3
Rozbeh Chiryai Sharahi

### 💻 Environment

...

...

_No response_

fzlzjerry:
Fixes #5036

Service log rows now resolve field settings from their source container instead of an empty map. Replicas with the same image (without tag) and command still share settings. Different images or commands remain independent.

- Regression coverage includes drawer toggles, new messages, late metadata, shared replicas and search.
- Checks: 300 frontend tests, typecheck, production build, Staticcheck and all 17 Go test packages with `-race` as non-root.
- Browser: desktop and mobile passed toggling, source isolation, container-view sharing and persistence after reload, using real Docker logs.
- The browser setup used service-labelled containers in server mode, not a native Swarm cluster.

A separate test-only commit fixes a pre-existing timing-dependent recent-container fixture.

### DIFF
--- a/assets/auto-imports.d.ts
+++ b/assets/auto-imports.d.ts
@@ -424,6 +424,7 @@ declare global {
   const useWindowFocus: typeof import('@vueuse/core').useWindowFocus
   const useWindowScroll: typeof import('@vueuse/core').useWindowScroll
   const useWindowSize: typeof import('@vueuse/core').useWindowSize
+  const visibleKeysForContainer: typeof import('./composable/storage').visibleKeysForContainer
   const watch: typeof import('vue').watch
   const watchArray: typeof import('@vueuse/core').watchArray
   const watchAtMost: typeof import('@vueuse/core').watchAtMost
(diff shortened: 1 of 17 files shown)


amir20:
Fixes more instances

PR #5109, opened 2026-09-11
fix(podman): derive a stable host id instead of trusting /info

### LINKED ISSUE #5098: Podman: `/info` ID is regenerated on every call, so the documented `engine-id` workaround for "host not found" is a no-op

- [x] Completed

### How is Dozzle deployed?

Agents

### 📦 Dozzle version

v11.0.0

Podman quadlet

The Podman guide (docs/guide/podman.md) recommends creating a stable `/var/lib/docker/engine-id` file to avoid `host <uuid> not found` errors, citing "Podman doesn't create an engine-id like Docker." This does not fix the issue, because Podman's compat `/info` endpoint never reads any file for that field in the first place — it generates a brand new random UUID on *every single call*, unconditionally:

https://github.com/containers/podman/blob/main/pkg/api/handlers/compat/info.go#L87

```go
ID: uuid.New().String(),
```

This is unconditional (no fallback-if-file-missing branch) and unchanged between the Podman version I tested (5.4.2) and current main, so mounting any file at any path has no effect on any Podman version, past or future, unless Podman itself changes this.

Reproduction

Query the same rootless Podman socket Dozzle mounts, twice, a few seconds apart:

```sh
curl -s --unix-socket /run/user/<uid>/podman/podman.sock "http://d/v1.40/info" | jq .ID
curl -s --unix-socket /run/user/<uid>/podman/podman.sock "http://d/v1.40/info" | jq .ID
```

Each call returns a different UUID, with or without any engine-id file mounted at /var/lib/docker/engine-id.

Actual mechanism causing "host not found"

NewClient (internal/docker/client.go) sets host.ID = info.ID unconditionally, and this happens once per process (agent or local hub
client), caching whatever random ID Podman happened to return at that moment for that process's whole lifetime. The hub
(RetriableClientManager) likewise records a client's ID only at first successful connect and never re-reads it for a host it already considers connected (RetryAndList only retries hosts already in failedAgents).

So on Podman:
- One agent's identity stays stable only as long as its own process doesn't restart.
- The hub never notices an agent silently getting a new ID after a restart it didn't itself observe as a disconnect - it just logs host <uuid> not found / error while trying to find container indefinitely for the stale ID, with no self-healing short of restarting the hub itself.

### 💻 Environment

- Dozzle: v11.0.0 (also confirmed on v10.10.0's client code, same info.ID usage)
- Podman: 5.4.2, rootless, Quadlet-managed
- Deployment: agent mode, ~15 hosts, one hub

host <uuid> not found / error while trying to find container

_No response_

amir20:
Fixes #5098. The reporter is right on both counts, and there are two separate bugs behind that issue.

## The docs were wrong

Podman's Docker compat `/info` sets `ID: uuid.New().String()` unconditionally and reads no file on the way. Same line at v2.0.0, v4.0.0, v5.4.2 and main, so the `/var/lib/docker/engine-id` workaround this repo has recommended for years has never done anything on any Podman version.

It isn't a stubbed-out bug upstream either. Podman is daemonless, so there's no process to own an engine identity and no file to persist one. `libpod/define/info.go` tracks no machine or engine ID at all. The compat handler has to fill a required Docker field, so it generates a throwaway.

## The real bug was ours

`NewClient` caches `info.ID` once per process, so a Podman host's ID lived only as long as the process that first asked for it.

The hub then kept routing to an ID nothing answered to:

- `m.clients` is keyed by the ID seen at first connect and never re-keyed
- `RetryAndList` only revisits endpoints that never connected, so a live-but-restarted agent is never re-read
- `Hosts()` hands the UI the *live* ID fetched from the agent

After an agent restart those two disagree, and every `Find` for that host fails with `host not found` until the hub itself is restarted. Docker just hides this because its ID is stable.

## What this changes

**Derive the ID on Podman** from the hostname plus the storage graph root, the two things Podman does report consistently. Graphroot is what keeps two rootless users on one machine apart, since they share a hostname but never a store. Docker, swarm and k8s keep exactly the IDs they already have; changing those would migrate every existing install for no gain.

It returns empty and falls back to the random ID when Podman reports neither value. Hashing two empty strings would hand every host in a fleet the same ID, and the duplicate-host check would then collapse them into one, which is worse than churn.

**Re-key the hub** when a client's reported ID stops matching its map key. This is the part that actually stops the permanent `host not found`, and it works for agents too old to have the derived ID. It refuses to re-key when the new ID already belongs to a different live client, since that's the duplicate-host case rather than a restart, and warns instead of evicting a host that's working fine.

**`DOZZLE_HOST_ID`** for the one case the derivation can't solve: two hosts sharing both a hostname and a storage root. Validated against `[A-Za-z0-9_.-]+`, since it lands in a path segment of every `/api/hosts/{host}/...` route. Swarm mode deliberately ignores it; identity there is the swarm node ID.

**A `replacesId` field** on re-keyed hosts, so an open tab drops the stale sidebar entry instead of listing one machine twice.

Also fixes a pre-existing unsynchronized read of `failedAgents` in `Hosts()` that `-race` would have caught once that path got busier.

Docs: the engine-id advice is gone from `README.md`, `podman.md` and `faq.md`, replaced with the actual mechanism and a two-line curl that shows the ID changing. All four locales updated and re-stamped.

## Host ids are resolved by an injected resolver

Later commits, in response to review. The id used to be built by a ladder of assignments in `newClient` where later lines silently won, with k8s doing its own thing in another file.

`container.HostIDResolver` is now chosen once at startup and injected into whatever connects to an engine. A client reports what it found and is told what it is talking to:

```go
host.ID = hostIDs.Resolve(container.EngineIdentity{
    Runtime:     runtime,
    EngineID:    info.ID,
    SwarmNodeID: info.Swarm.NodeID,
    Hostname:    info.Name,
    StorageRoot: info.DockerRootDir,
    Fallback:    host.ID,
})
```

No precedence, no branch, no knowledge of what any of it is worth. Two implementations: `StaticHostID` for `DOZZLE_HOST_ID` and `DerivedHostID` for working it out. The Podman derivation moved with them, since it is a rule about identity rather than anything to do with the Docker API. k8s builds an `EngineIdentity` from its node and resolves through the same seam, so `DOZZLE_HOST_ID` works there too.

Wiring it explicitly made a trap visible that was implicit before. The override names the host *this process* runs on, so only the local client gets it; handing one `StaticHostID` to several `--remote-host` engines would give them all the same id and collapse them into a single host. The composition root now says that in one place instead of it being an accident of which constructor took the parameter.

The tests split along the same line: `internal/container` covers what the id should be, and `internal/docker` covers only that the client describes its engine accurately and uses the answer.

That also turned up dead code: `ParseConnection` sets `Host.ID` from the remote URL and the client always overwrote it, including with `""` when the engine was unreachable. It is now `Fallback`, so a remote host Dozzle cannot reach keeps an id that routes instead of an empty one that cannot. That is a small behavior change beyond the issue, called out here rather than buried.

## Notes

- Podman host IDs change once on upgrade, so bookmarked `/host/<id>` URLs and pinned state break for existing Podman users. Worth a release note.
- An already-open tab heals on reload rather than instantly. `Hosts()` is the repair point and it runs on page load, not on a timer. Happy to add detection on the events-stream reconnect too if that's too loose.

## Testing

22 new tests covering the resolver, the derivation (stability across restarts, rootless separation, the empty-info fallback, override and swarm precedence) and the re-key (happy path, subscriber notification, unavailable hosts left alone, refusing to clobber another host, stable hosts untouched). Full Go suite passes under `-race`; frontend tests and typecheck green.

https://claude.ai/code/session_01Csk291RcrUyML7vPEWaRyF


FabienYt:
Thanks a lot for the quick turnaround on this, and for Dozzle in general, really appreciate the work.

I tested amir20/dozzle:pr-5109 against the exact scenario from my report: one hub in agent mode with ~14 remote Podman agents over TLS, plus one local socket. Steps:

1. Rolled back my own workaround (forcing a hub restart right after every agent redeploy) and bumped both hub and agent images to pr-5109.
2. Fresh hub restart, confirmed clients: 15 in the log, no errors.
3. Restarted one agent's container independently of the hub (simulating an image bump, a unit bounce, or a reboot outside of a coordinated deploy). This is the exact trigger described in the issue.
4. Watched the hub for about 5.5 minutes: no more host <uuid> not found spam, which is an improvement over main, but the hub's outbound TCP connection to that agent (port 7007) was never re-established on its own. I confirmed this with ss on the hub host: the connection just stayed absent the whole time.

Looking at the merged code (retriable_client_manager.go (https://github.com/amir20/dozzle/blob/010b5fc189faa1c74b31ff1cde08de0368c9ce06/internal/support/docker/retriable_client_manager.go)), rekey() only runs inside Hosts()/RetryAndList(), and those are only invoked by an external caller such as the web UI listing hosts. There is no ticker or background goroutine that calls them on its own. So the fix works well when the dashboard is already open when an agent restarts, since it then self-heals on the next host-list refresh instead of being stuck forever, but for a deploy where nobody opens the UI right afterward, the hub still won't reconnect to that agent until something triggers Hosts().

Is that the intended scope, relying on the UI's own polling to eventually reconcile, or would a periodic reconciliation independent of a caller be worth adding? Happy to test again if there's a build with that.

Thanks again!

amir20:
That's the intended scope, yeah. The hub connects to agents lazily. It has no reason to hold a connection open when nothing is asking it for anything, so an absent socket with no UI open isn't a symptom. The first request that needs that host dials it again, and fails after the configured timeout if the agent is still down.

The bug in #5098 was that the reconnect happened and still didn't help, because the hub kept routing to the old id forever. That part is fixed.

You're right that `rekey()` only runs from `Hosts()` though, and I don't want to add a ticker for it. A background reconciler would dial every agent on a timer whether or not anyone is watching, which is a lot of connection churn for a fleet of 14 to fix something a page load already fixes.

The case I do want to cover is a dashboard that's already open when an agent restarts. Right now it heals on reload rather than on its own. The events stream already reconnects every few seconds after a drop, so repairing there gets you live recovery without a timer. I'll open a follow-up for that.

PR #5124, opened 2026-09-12
feat: add oidc auth provider that reads users and roles from the token

### LINKED ISSUE #5121: Read all user information from OpenID Connect token

Thank you for integrating OpenID Connect directly into Dozzle!

The current implementation still requires a `users.yml` file for maintaining users. I would like to be able to manage user access to Dozzle directly in my IdP by assigning roles, groups or other attributes to them and including these as claims in the OpenID Connect Access Token or ID Token, so Dozzle can extract the necessary information directly from there.

- `DOZZLE_AUTH_OIDC_ROLES_CLAIM` points to an array claim in the token containing the roles of a user
- `DOZZLE_AUTH_OIDC_FILTERS_CLAIM` points to an array claim in the token containing the filters of a user

Given that a user’s token contains

```json
  "resource_access": {
    "dozzle": {
      "roles": [
        "shell",
        "actions"
      ],
      "filters": [
        "label=com.example.app"
      ]
    }
  },
```

`DOZZLE_AUTH_OIDC_ROLES_CLAIM=resource_access.dozzle.roles` and `DOZZLE_AUTH_OIDC_FILTERS_CLAIM=resource_access.dozzle.filters` should be set.

- A user’s username, full name and email address are read from the standard claims `preferred_username`, `name` and `email` respectively (alternatively, custom claims may be configurable)
- If a user does not have a role claim set or the array is empty, access to Dozzle is denied
- For storing user information at `/data/<username>/profile.json`, the `sub` claim may be used instead to uniquely identify the user

_No response_

amir20:
Adds --auth-provider oidc, where the identity provider is the user database
and users.yml is never read. Roles come from the first of dozzle_roles,
resource_access.<client-id>.roles and roles that exists in the ID token or
userinfo, filters from the matching filters claims, and both can be
overridden with --auth-oidc-roles-claim and --auth-oidc-filters-claim.

A login with no roles claim, or an empty one, is refused and the log names
the paths tried. A claim that names no Dozzle role signs in with no
privileges, like roles: none. sub keys the user and the profile directory,
the email does not have to be verified, and a picture claim feeds the
avatar by redirect. Roles and filters ride in Dozzle's session JWT and are
re-parsed per request, so a change at the IdP lands at the next login.

The OAuth code exchange is shared with simple auth through oauthFlow, so
the two providers differ only in what happens to the identity. Under oidc
there is no password form or POST /api/token, the GitHub flags are a
startup error, and --auth-logout-url is honored after the session is
cleared.

Docs get a new OpenID Connect page and updates to the authentication
overview, the GitHub & OIDC page and the env var table, translated into
de, fr, es and zh.

Closes #5121

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014X3D6uNjy6WjMU849WfRxm

### DIFF
--- a/assets/components/shell/Links.vue
+++ b/assets/components/shell/Links.vue
@@ -61,7 +61,7 @@
             </div>
           </div>
 
-          <template v-if="config.authProvider === 'simple' || config.logoutUrl">
+          <template v-if="hasSession || config.logoutUrl">
             <div class="bg-base-content/10 my-1.5 h-px"></div>
             <button
               @click.prevent="logout()"
@@ -79,6 +79,10 @@
 <script lang="ts" setup>
 const { mounted: cloudSurfaceMounted } = useCloudSurface();
 const { logoutUrl } = config;
+
+// simple and oidc hold a session cookie Dozzle issued, so logging out means
+// clearing it. Forward proxy has none: the logout URL is the whole logout there.
+const hasSession = config.authProvider === "simple" || config.authProvider === "oidc";
 
 // The bell is the only place that watches for a fire the reader has not seen, so
 // it owns the polling. Without a cloud link there is nothing to remember and
@@ -101,13 +105,17 @@ watch(visibility, (state) => {
 });
 
 async function logout() {
-  if (logoutUrl) {
-    location.href = logoutUrl;
-  } else {
+  if (hasSession) {
     await fetch(withBase("/api/token"), {
       method: "DELETE",
     });
+  }
 
+  // Under oidc the URL is where to go once the session is gone, usually the
+  // issuer's own logout, so it runs after the DELETE rather than instead of it.
+  if (logoutUrl) {
+    location.href = logoutUrl;
+  } else {
     location.reload();
   }
 }
--- a/assets/stores/config.ts
+++ b/assets/stores/config.ts
@@ -10,7 +10,7 @@ export interface Config {
   hostname: string;
   mode: "server" | "swarm" | "k8s";
   hosts: Host[];
-  authProvider: "simple" | "none" | "forward-proxy";
+  authProvider: "simple" | "none" | "forward-proxy" | "oidc";
   oauthProviders?: { name: string; loginUrl: string; icon: string }[];
   passwordLogin?: boolean;
   logoutUrl?: string;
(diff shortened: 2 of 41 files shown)


marvinruder:
@claude Something doesn't work yet, the OIDC provider redirects to `/api/auth/callback`, but that endpoint only responds with a 302 to `/`, which in turn responds with 307 to `/login?redirectUrl=%2F` as if I am not logged in yet. The `/api/auth/callback` does not set any cookies. The Dozzle logs appear to indicate successful authentication:

```json
{"level":"info","version":"pr-5124-0786905","user":"<UUID>","login":"<username>","claim":"resource_access.dozzle.roles","roles":["none"],"time":"2026-09-12T20:30:07Z","message":"Token created"}
```

amir20:
I have never tried tagging claude. :) What OIDC are you using? I wonder if I can set it up locally to test it.

marvinruder:
> I have never tried tagging claude. :)

Haven’t worked with Claude in GitHub before, just wanted to see if / how I can let Claude improve the PR based on my feedback.

> What OIDC are you using? I wonder if I can set it up locally to test it.

I use Keycloak. I just noticed that when I exclude the `picture` claim everything works fine. My `picture` claim holds a `data:` URL so my ID token is somewhere between 4 KB and 8 KB. That setup works fine with other clients, but I guess Dozzle cannot handle it right now. Interestingly I do not get an error log message showing what went wrong where, I just don’t get a cookie.

amir20:
I just brought up keycloak and it worked as well. 

I'll see what is happening with `picture`. It might be a parsing issue.

amir20:
found it. dozzle was copying the `picture` claim into its own session cookie, and a data: URL pushes that past the browser's 4KB cookie limit so the browser silently drops it. pushed c3f96eab: pictures are only kept when they're an https URL, and a session that's still too big fails with a log line instead of silently. can you try `amir20/dozzle:pr-5124` once CI rebuilds?

marvinruder:
Yes, works fine now!

amir20:
Alright. Merging. It shold update the docs after I merge. Then you can try `master` for a while. If everything is ok then I can release.

PR #5128, opened 2026-09-13
fix(data): keep symlinks and umask in atomic writes

amir20:
follow-up to #5127 review.

- symlinked `notifications.yml` / `cloud.yml` were replaced by a regular file on save. the path is now resolved first, so the link stays and its target gets replaced.
- new files were always chmod'd 0644, ignoring umask. a file that doesn't exist yet is now created directly with O_EXCL, so umask applies like it did with os.Create. there's no old version to lose on first create anyway.
- rename and temp file fallbacks are tested by swapping in failing funcs, so they're exercised even when tests run as root.
- failed dir fsync is logged at debug.

https://claude.ai/code/session_01Ny9NZ63Q9GtmJYEEStvkAi


amir20:
closing, this is getting too involved for how rare the problem is. replacing the rename approach with a plain in-place write (render first, fsync, per-path lock).

PR #5129, opened 2026-09-13
fix(data): replace atomic rename with a plain synced write

amir20:
the temp file + rename from #5127 broke symlinked configs and ignored umask, and fixing those in #5128 kept turning up more edge cases. too much for a problem nobody reported.

back to os.Create, keeping the useful parts:
- render into memory first, so a failing encoder can't leave a truncated file
- fsync after writing
- per-path lock, so concurrent saves (SaveNotifications has no lock of its own) can't interleave

mounts, symlinks and permissions behave the same as before #5127.

https://claude.ai/code/session_01Ny9NZ63Q9GtmJYEEStvkAi


amir20:
switched to a single global lock, the per-path map never shrank. leaving the truncate window as is, that's the tradeoff this PR makes. skipping the dir fsync on new files too, the code before #5127 never had it and it only matters for a crash right after first creation.

PR #5145, opened 2026-09-13
feat(icons): Add icons for more services

achalker:
Adding additional icons, including some generic ones the could be of use via manual container label tagging


amir20:
Today I learned about what's up docker 😀

PR #5154, opened 2026-09-14
feat(hosts): select a host by clicking its name on the dashboard

d-shiri:
The host cards on the dashboard show a name, container count and meters, but the name is plain text. Switching hosts means going back to the sidebar and clicking the host there.

Clicking the name now sets `sessionHost`, the same as clicking its row in the sidebar. The nav slides to that host's containers and the main pane stays where it is.

It does not link to `/host/[id]`. That route is the merged stream, and the sidebar's merge button already covers it.

One file, `assets/components/hosts/HostCard.vue`. No new strings, no new route.

https://claude.ai/code/session_01NCY7GHwmakhQQLigQ7bMwu


amir20:
🇮🇷🦁

PR #5166, opened 2026-09-15
feat: add 'fail' level to level_guesser.go

BLNJ:
The standard C# ASP.Core Logger has a "fail" log level, that is comparable to "fatal".
image

This PR adds the log level to the logguesser.
Honestly I have no idea what Im doing


BLNJ:
@amir20 thank you so much for expanding on what I tried to do :-)

PR #5200, opened 2026-09-20
feat(charts): read out the hovered bar on host and container stat charts

amir20:
Hovering a host card chart now swaps the headline to that bar's value and the subline to when it was. The container table's sparklines read out the same way. `BarChart` was already emitting `hoverValue`, `MetricCard` just ignored it.

While wiring that up, the hover handling moved into `BarChart` instead of being reassembled by every consumer:

- the `opacity-70 hover:opacity-100` affordance was copy-pasted into three call sites and missing from a fourth, which is why the host card charts looked dead
- each consumer hung its own `@mouseleave` on whichever ancestor it owned, so `BarChart` emits `hoverEnd` now
- the hit test asked every bar for its rect on each pointer move. Bars are `flex-1` siblings with a fixed gap, so one rect on the container plus arithmetic gives the same index

A hovered bar also draws a 1px guide on its column, because the opacity shift is invisible at 3px wide and at `h-4` in a table row.

## The padding was being read as data

`statsHistory` is padded at the front so a chart keeps its width while real samples scroll in. That padding is not a measurement, and everything downstream was treating it as one. On a freshly started Dozzle I measured 66 of 100 bars on the host CPU chart and 28 of 43 in a table cell as padding, which meant:

- a long flat stretch that reads as an idle container
- `avg` divided by samples nobody took (one card showed `avg 414.9 KB` beside `pk 7.2 MB`)
- and the new hover would have stated `0.0%` at a timestamp with total confidence

`Container` now tracks how many samples are real. Charts draw the padding as a faint track, and it is excluded from the averages and from the hover.

Container CPU series also carry the per-core value now, so the hover matches the number printed next to it instead of raw docker percent.

## Testing

442 frontend tests pass, plus typecheck and lint. New coverage for the padding, the guide, and the `MetricCard` readout. Verified by hand against a running instance.


amir20:
Pushed fixes for two of the three.

**Boundary bucket dilution** (greptile, P1): real, fixed. A bucket straddling the boundary counted as sampled but still averaged the padding's zeros, so the oldest readable bar reported below what was measured. It now averages only the real points. Test asserts 90 where the old code gave 72.

**Stale sampled flags** (greptile, P1): I tried fixing this and it caused a worse bug, so I've deliberately left it. Watching the boundary and recalculating on a bucket crossing re-buckets the series against data that has only moved a fraction of a bar, and every bar jumps instead of the chart scrolling left by one. That is what `changeCounter` exists to prevent: a recalculation is only phase-stable every `bucketSize` ticks. The cost of not watching it is that a boundary bar can stay faint for up to `bucketSize` ticks, which is the same cadence at which it would be redrawn anyway. There's a comment in the file now saying so, and a regression test that fails if anyone adds the watch back.

**`Math.max` vs `Math.min`** for the aggregate sample count: I think `max` is right, though the code deserved a comment and now has one.

The two cases differ in why a history is short. If Dozzle just started, every container's history is equally short, so `max` and `min` are the same number and the choice doesn't matter. They only diverge when a container was created recently, and there the padding is not "no data yet", it's "this container did not exist", so zero is the correct contribution to a host total. `min` would also mean a single newly created container blanks the sampled region of the whole host chart, which throws away real history from everything else on the box.

What is genuinely missing is that `statsHistory` can't currently distinguish "not running yet" from "not observed yet". Worth splitting if it ever bites, but `max` is the better of the two behaviors today.

PR #5203, opened 2026-09-20
perf(nav): mount nav popup contents on first open

amir20:
The sidebar renders a `Popup` per container, and every one mounted its panel contents with the row. On a host with 76 containers that's 76 live `ContainerPopup` trees for panels nobody has opened, each holding two `ContainerStatCell` charts.

The charts aren't idle either: `BarChart`'s watcher reads `chartData`, so `ContainerStatCell` mapped a 300-entry `statsHistory` into 300 fresh objects once a second, per chart, entirely offscreen.

Contents now mount on first open and stay mounted. A row hovered once tends to be hovered again, and remounting per hover would redo the work this is avoiding. The mount is awaited before `showPopover()` because `useAnchoredPopover` measures and positions the panel immediately after, and an empty box positions wrong.

## Measured

Same page, same viewport, 76 containers:

| | before | after |
|---|---|---|
| DOM nodes | 5,509 | **3,074** |
| chart instances | 184 | **32** |
| `Popup` panels holding content at rest | 76 | **0** |

The opened panel is unchanged: 280x214 at the same coordinates, `visibility: visible`, with its two charts mounted on demand.

`Popup` has exactly one consumer (`HostMenu`), so the blast radius is that one surface.


amir20:
Added `Popup.spec.ts`, four cases. jsdom has no popover API so `showPopover` is stubbed to record whether the content was in the DOM when it ran, which is the ordering worth locking in. Confirmed the suite fails if the content goes back to being mounted with the row.

On `hasOpened` never resetting: that's intended, and the erosion is bounded by how many rows someone actually hovers rather than by the container count. A panel is a name, a few rows of text and two sparklines, so a session would have to hover a large fraction of the sidebar to get back to where it started, and at that point they're all panels the user has demonstrably wanted. Unmounting on close would trade that for re-running the mount on every hover of a row someone is scrubbing up and down, which is the case this is trying to make cheap.

Happy to switch it to unmount-on-close if you'd rather have the flat ceiling.

PR #5204, opened 2026-09-20
fix(charts): follow-ups from the hover readout review

amir20:
Follow-ups from the review of #5200. Six findings, three of which let the readout state something false.

## The readout could report padding as a measurement

`HostCard`'s backfill replaces the whole series, but its length stays 300, so nothing in the chart notices: it keeps the old bars *and* their old `sampled` flags until the next scheduled recalculation. A container starting or stopping meant hovering reported padding as a real value, with a timestamp. That's exactly what those flags exist to prevent.

`MetricCard` renders the `BarChart` itself, so the parent had no way to reach it. It forwards `recalculate()` now, the way `MultiContainerStat` already did.

## The hovered value froze

The pointer sits over a fixed column while the series scrolls underneath it. The value is now re-reported from whatever is drawn in that column on each update. Previously it froze on the sample under the pointer when it stopped moving, then went quietly wrong at the next recalculation.

It reports what is drawn, so the readout and the bar it points at can't disagree.

## Unmounting is a silent way to leave a chart

Switching the container table to progress mode while hovering unmounts the chart with no `mouseleave`, so `hover-end` never fired and the cell showed the frozen hovered number instead of the live average, indefinitely. `BarChart` releases the hover on unmount now, which covers every consumer.

## Smaller

- the guide clamps its index against the current bar count, so a resize can't throw it outside the chart
- the guide sits on the bar's centre rather than half a gap to its right
- the guide and the hit test read one width source, instead of `useElementSize` (content-box) and `getBoundingClientRect` (border-box) separately
- dropped a `watch(() => chartData.length)` that could never fire, since the length is pinned at 300

## Testing

453 tests, typecheck and lint green. Three new tests for the behavioural fixes, each confirmed to fail without its fix.


amir20:
Good catch, fixed. `reportHovered()` was only reachable from the `chartData` watcher, which left out the two paths this PR added: a parent's forced `recalculate()` (the exact scenario the PR exists for) and a plain resize.

Moved it into the two functions that actually mutate the bars, `recalculate()` and `updateLastBar()`, so every caller gets it rather than each call site remembering.

One behaviour note on the resize case. When the re-bucket leaves the pointer past the end of a shorter series, the hover is released instead of clamped. There is no `clientX` to re-run the hit test with until the pointer moves again, so naming a different bar would be a guess. The guide's own clamp stays as cheap insurance against ordering surprises, but it is no longer the thing doing the work.

Two tests, both confirmed to fail with the previous placement:
- a forced `recalculate()` re-reports the hovered bar. It swaps the series while keeping the length at 300 and reusing the last entry, so `chartData.at(-1)` never changes identity and the watcher stays blind, which is exactly what HostCard's backfill does.
- a resize releases a hover that is no longer on a bar.

PR #5207, opened 2026-09-20
perf(charts): draw a chart as one path instead of a div per bar

amir20:
Rebased onto master now that #5206 is merged, so the diff is just this change.

A sparkline in a table cell was 75 divs, each carrying a class list, a `--height` custom property, `will-change: height` and `contain: layout`. At 100 containers that is ~15,000 of them and about 90ms a second spent restyling. #5206 took the reactivity cost out of the stats path and left this as what was left.

Each layer is now one `<path>` with a subpath per bar, so a chart is three nodes however many bars it draws, and a tick writes one `d` attribute.

Bars are stroked vertical lines rather than rects. `stroke-linecap="round"` gives the rounded top for free, the matching bottom cap falls outside the viewport and gets clipped, and the geometry stays one `M x yV y` per bar.

## Numbers

Real Chromium, ms per stat tick, median of 3 runs, round-robin between the three builds so drift hits all of them equally.

| containers | pre-#5206 | +#5206 | +this | DOM nodes |
| --- | --- | --- | --- | --- |
| 10 | 15.5ms | 8.3ms | 6.0ms | 1,472 -> 132 |
| 25 | 36.0ms | 21.5ms | 15.1ms | 3,647 -> 327 |
| 50 | 80.1ms | 40.4ms | 27.6ms | 7,352 -> 652 |
| 100 | 162.0ms | 88.2ms | 69.9ms | 14,777 -> 1,302 |

Markup for a 100 container table drops from ~2.5MB to ~670KB.

Screenshotted both builds side by side at 3x. The rendering is indistinguishable: same bar widths, same rounded tops, same colours, same density.

## Call sites

`barClass` is a text colour now, not a background, because bars stroke in `currentColor`. The four call sites pass `text-primary` where they passed `bg-primary`, all updated here.

Per-bar `hover:opacity-100` has no element left to hang on, so the hovered bar is redrawn opaque as a third path.

## Testing

All 14 BarChart tests pass. They read bar geometry back out of the path `d` instead of off each div, so they still assert what is drawn rather than internal state. 465 tests green, typecheck, lint and production build clean.

No visual snapshots to regenerate: `e2e/visual.spec.ts` only captures `[data-testid="navigation"]`, and the mobile stats specs assert text, not chart pixels.

https://claude.ai/code/session_01P1pZSvMJ6RrgXKU7yaz5He


amir20:
Good catch on `usableHeight`, fixed in c54b121. The cap allowance is now `min(barWidth / 2, height / 2)`, so the subtraction can never eat the whole height.

One correction though: it's not only reachable in theory. `RailMetrics.vue` feeds `body.points` straight from the cloud API for whatever window is selected (1h/6h/24h/7d), so that call site is a server-supplied series of arbitrary length, not the fixed 300-item `statsHistory` the stats charts pass. So the invariant you suggested documenting doesn't actually hold today.

Added a test that fails without the clamp: 4 points in a 16px chart, asserting every bar has non-zero height.

---

PR #5208, opened 2026-09-20
fix(store): drop containers a host no longer lists

amir20:
Off master rather than stacked on #5207, since it only touches `assets/stores/container.ts` and neither perf PR goes near it. Happy to restack if you'd rather keep them in one line.

Nothing ever removed a container from the store. `destroy` only marks one `"deleted"`, which hides it from every list but keeps the object, and a container that disappears while the tab is asleep raises no event at all. The only code that shrank `containers` was the reconnect handler, so a tab left open on a host with any churn held every container it had ever seen until someone reloaded.

## Cost

Each dead container keeps its own stats window. Retained heap per container, measured in Chromium with precise memory info:

| lifetime before death | stats kept | bytes each | 1,000 of them |
| --- | --- | --- | --- |
| ~10s (a CI job) | 10 | 3.7 KB | 3.6 MB |
| ~1 min | 60 | 7.1 KB | 6.9 MB |
| 5 min or more | 300 | 24.8 KB | 24.2 MB |

Driving the real store through 40 rounds of 25 containers that start, run for a minute and get removed:

```
                        before   after
containers held          1,001       1
containers alive             1       1
heap growth              7.5 MB  0.6 MB
```

## The fix

`containers-changed` is authoritative for every host it names, so a container of a named host that is missing from the payload is gone and gets dropped.

Scoped by host, never wholesale. Only the payload sent on connect covers everything. The others carry one host's list after a start, a rename or a stale-host repair (`refreshHost` in `internal/web/events.go`), and treating one of those as the whole world would evict every other host. There's a test for exactly that.

The comment on `visibleContainers` already described this behaviour ("lingers as deleted until the next full list"). It just was not implemented.

## Behaviour change worth a look

A destroyed container you are currently viewing now flips to the Not Found page once a list covering its host arrives, instead of staying on screen indefinitely. `pages/container/[id].vue` already handles the container going missing, and the same thing already happened on any reconnect, so this is not new ground. But it is a real change: on a busy host it lands within seconds rather than never. Say the word if you would rather keep the logs on screen and bound the growth some other way.

## Tests

Four tests driving the real store through a stubbed EventSource. Two fail on the current code (accumulation, and the multi-host eviction the scoping prevents); the other two pass either way on purpose, guarding against a fix that prunes too eagerly: an empty payload names no host so it must drop nothing, and a container that survives a list must keep its identity along with the stats history and EMA it has built up.

468 tests green, typecheck, lint and production build clean.

While I was in here I checked the other module-level caches that grow with container churn. `composable/containers/imageUpdate.ts` is already bounded at 200 entries with this exact reasoning written on it, so the store was the outlier rather than the pattern.

https://claude.ai/code/session_01P1pZSvMJ6RrgXKU7yaz5He

---
_Generated by [Claude Code](https://claude.ai/code/session_01P1pZSvMJ6RrgXKU7yaz5He)_


amir20:
Went through all three.

**Empty hosts remain retained** (greptile P1 #1). Real, and deliberate. An empty payload names no host, so there is nothing it can be authoritative about, and treating it as global would evict every other host's containers. That is the exact case the multi-host test covers, so I would rather keep the gap. The consequence is bounded: one host's already-dead containers linger until any non-empty list for that host arrives, which the next start or rename produces. It is not the unbounded growth this PR is about, and nothing accumulates across generations.

**Older snapshots remove live containers** (greptile P1 #2). Real, and it is a backend bug rather than a frontend one. `events.go` already guards this race on the error branch ("a start/rename may have repaired this host while the retry was in flight") and does not on the success branch, so a repair that finishes after an inline refresh overwrites it with an older list. Sending that as its own PR.

Worth noting it already misbehaves today, before this change: the same branch calls `setVisible(refresh.host, refresh.containers)` with the stale set, so the container that started is gated out of stats and events either way. Today that leaves a frozen row that never updates; with this PR it disappears instead. Both self-heal on the next list for that host, and neither is good, which is why the fix belongs in the repair handler.

**Pinned columns vanishing** (claude[bot]). Correct, and worth saying that it already happens: `onopen` clears `containers` on every reconnect, so a pinned column for a destroyed container disappears then too. This makes it happen sooner rather than introducing it. `pinnedContainerIds` keeps the id either way, so nothing is lost from the URL. Whether a vanished column deserves a placeholder is a product call I would rather not make inside a leak fix. Happy to add one if you want it.

---

PR #5212, opened 2026-09-20
fix(events): read Podman health_status from actor attributes

### LINKED ISSUE #5211: Health status stays "starting" forever with Podman: health_status event carries the value in a separate field

- [x] Completed

### How is Dozzle deployed?

Agents

### 📦 Dozzle version

v11.1.0

NA

With Podman as the container engine, the container health badge stays on "starting" and never becomes "healthy", although the container is healthy (`podman ps` shows `(healthy)`, `podman inspect` reports `Health.Status: healthy`).

The badge is frozen at whatever value Dozzle read the first time it listed the container. It only ever updates on a fresh listing, never on health events.

### To reproduce

1. Run Dozzle (v11.1.0) against a rootless Podman socket, directly or through an agent.
2. Run any container with a `HEALTHCHECK` / `--health-cmd` that succeeds.
3. Wait for the health check to pass. `podman ps` shows `(healthy)`.
4. Dozzle still shows "starting".
5. Restart the agent: containers already healthy at that moment flip to "heal The agent's own container, and any container restarted afterwards, stay on "starting".

### 💻 Environment

- Dozzle v11.1.0, hub + remote agents (agent mode)
- Podman 5.4.2, rootless, Docker-compatible API socket (`podman.sock`) mounted read-only
- Debian, Linux amd64 and arm64
- Also seen on v10.8.0 and v10.9.0

Docker encodes the new health state in the event action itself:

`    Action: "health_status: healthy"`

Podman's Docker-compatible events endpoint sends the bare action and puts the value in a separate Podman-specific field:

```
    Action:       "health_status"
    HealthStatus: "healthy"
```

`podman events` shows the same thing:

`    container health_status <id> (image=..., name=..., health_status=healthy, health_failing_streak=0, ...)`

An event is emitted on every health check run (~30 s here), so the events do arrive; they just don't match the `health_status: healthy` / `health_status: unhealthy` string Dozzle looks for, so no transition is ever seen.

### Expected behavior

The badge follows the container's health, as with Docker. Ideally Dozzle handles both encodings: `Action == "health_status: <state>"` (Docker) and `Action == "health_status"` with the state in the `HealthStatus` attribute (Podman).

_No response_

team-humaki:
Fixes #5211.

Docker encodes health in the event action (`health_status: healthy`). Podman's Docker-compatible events endpoint sends the bare action `health_status` and puts the value in actor attributes (`health_status=healthy`), so Dozzle never saw a transition and the badge stayed on whatever the first list returned.

Handle both encodings: parse `health_status: <state>` as before, and when the action is bare `health_status` read `healthStatus` / `health_status` / `HealthStatus` from actor attributes. Apply that in the docker event mapper, the container store, the SSE health event, and notification matching.

`go test ./internal/container/ ./internal/container/docker/ ./internal/notification/ ./internal/web/` green.


amir20:
did you test this against an actual podman socket? I don't think it can work.

podman doesn't put the health status in actor attributes. `ConvertToEntitiesEvent` (`pkg/domain/entities/events.go`) builds attributes from `Details.Attributes` plus image, name, podId, containerExitCode, oomKilled, network, error. `HealthStatus` is a separate top-level field on the event struct. Same in 4.9, 5.4 and main. the journald backend agrees, `journal_linux.go:243` restores it as `newEvent.HealthStatus` and leaves `Attributes` as the container labels.

the `health_status=healthy` you see in `podman events` output comes from a printf off `e.HealthStatus` (`libpod/events/events.go:63`). it reads like an attributes dump but isn't one. I'm guessing that's where this came from, the issue quoted that same line.

so on a real podman host the action is bare `health_status`, attrs have no health key, `HealthStatusOf` returns false, and the badge stays frozen. the tests pass because they feed an attribute podman never sends.

there's a second problem underneath. `moby/api/types/events.Message` has no `HealthStatus` field, so even the right key gets dropped when the client decodes the stream. it never reaches `ContainerEvents`. getting at it needs either a `client.WithResponseHook` that rewrites the `/events` body, or an inspect when the bare action shows up.

the store/SSE/notification half is the right shape and `HealthStatusOf` is a decent seam. it just needs a producer that has the value.

team-humaki:
You're right — I was reading `podman events` printf output as actor attributes. `events.Message` has no HealthStatus field, so the value never reaches ContainerEvents.

This push inspects the container when the action is still the bare `health_status` (the producer you suggested). Docker's `health_status: <state>` path is unchanged. `starting` now reaches the store and SSE as well.

`go test ./internal/container/ ./internal/container/docker/ ./internal/notification/ ./internal/web/` green.

amir20:
OK. If https://github.com/amir20/dozzle/pull/5214 works better then that should be a better fix.

PR #5222, opened 2026-09-20
fix(pwa): stop iOS blurring the top of the installed app

amir20:
Replaces #5150, which moved the nav down 4px. That was too small by an order of magnitude, and it was treating a symptom.

`apple-mobile-web-app-status-bar-style="black-translucent"` plus `viewport-fit=cover` puts page pixels under the status bar. Since iOS 26 the system fills that inset with the Liquid Glass scroll edge effect, and the effect is a gradient, not a hairline at the edge.

Measured off a native 3x screenshot on a 16 Pro Max (1320px wide):

- The bee's antennae are smeared, the eyes are soft, the jaw facets at the bottom are pixel-crisp. Same on the search glyph (blurry circle, sharp handle tip) and the hamburger (top bar softer than the bottom one).
- Peak horizontal gradient across the nav row tops out around 120, against 200+ for both the iOS status bar text above it and the container bar text below it.
- It fades out around y=320 device px, so ~107pt, which is ~45pt past the 62pt safe area inset.

4px into a 45pt ramp does nothing. Clearing it with padding would need a nav twice as tall, and there is no CSS or meta opt-out for the effect (the `backdrop-filter` overlay trick stopped working in standalone), so the only lever is to stop drawing up there.

Dropping the meta means the web view starts below the status bar and `env(safe-area-inset-top)` reports 0. Every existing use of it (`pt-safe`, `--mobile-nav-offset`, the body padding) collapses to nothing on its own, so nothing else had to change. `viewport-fit=cover` stays for the landscape and home indicator insets. The value has also carried a removal warning in the Safari inspector since iOS 11.

The status bar is now painted from `theme-color`, which the markup was shipping empty until `App.vue` wrote the resolved theme into it after first paint. It gets the dark value the manifest already commits to for the same reason.

`bun run build` and `go test ./internal/web/...` pass. No snapshot regen needed, unlike #5150, since Chromium ignores the apple meta and safe area insets are 0 headless.

Needs a real device to confirm, image is `amir20/dozzle:pr-XXX` once CI builds it.

https://claude.ai/code/session_0125MhLLu2nxktCwQdfNrQus

---
_Generated by [Claude Code](https://claude.ai/code/session_0125MhLLu2nxktCwQdfNrQus)_


amir20:
This doesn't work. Correcting the PR body.

Measured the logo's top edge (10%-90% rise) on a device screenshot of `pr-5222-3ffb811` against the same edge on master:

- master: 4.3pt
- this PR: 5.7pt raw, but that screenshot is downscaled to 0.7x. The same SVG further down the page reads 2.4pt there, so the floor is 2.4, giving ~3.3pt of actual blur.

So ~4.0pt to ~3.3pt. That's noise, not a fix.

The reason is in the geometry. The blur band sits at the same place on screen in both builds, roughly 62pt (the safe area inset) down to ~100pt. It did not move down by 62pt when the web view moved down by 62pt. The effect is anchored to the safe area inset boundary and runs ~38pt deep whether or not page content draws under the status bar, so "stop drawing up there" does not get you out of it.

The other theory in the write-ups going around, that the system fills the band where WebKit cannot sample a solid color at the top edge, doesn't fit either. `--color-base-200` is `oklch(18% 0 0)` and the nav carries it, fully opaque.

Holding this until there's a way to verify a fix on a device rather than inferring one from screenshots.

---

